PR #19, opened 2015-03-04
Customizable summary field for JIRA tickets

jjsendor:
JIRA ticket summary can now be customized by adding an `alert_subject` that contains a custom summary, which can be further formatted using standard Python formatting syntax, e.g.:
```yaml
alert_subject: Issue {0} occurred at {1}
```

The arguments for the formatter will be fed from the matched objects.
The field names which values will be used as the arguments can be passed with `alert_subject_args`, e.g.:

```yaml
alert_subject_args:
- issue.name
- @timestamp
```

In case the rule matches multiple objects in the index, only the first match is used to populate the arguments for the formatter.

### DIFF
--- a/elastalert/alerts.py
+++ b/elastalert/alerts.py
@@ -78,6 +78,29 @@ class Alerter(object):
         """ Returns a dictionary of data related to this alert. """
         raise NotImplementedError()
 
+    def create_title(self, matches):
+        """ Creates custom alert title to be used, e.g. as an e-mail subject or JIRA issue summary.
+
+        :param matches: A list of dictionaries of relevant information to the alert.
+        """
+        if 'alert_subject' in self.rule:
+            return self.create_custom_title(matches)
+
+        return self.create_default_title(matches)
+
+    def create_custom_title(self, matches):
+        alert_subject = self.rule['alert_subject']
+
+        if 'alert_subject_args' in self.rule:
+            alert_subject_args = self.rule['alert_subject_args']
+            alert_subject_values = [matches[0][arg] for arg in alert_subject_args]
+            return alert_subject.format(*alert_subject_values)
+
+        return alert_subject
+
+    def create_default_title(self, matches):
+        raise NotImplementedError()
+
 
 class DebugAlerter(Alerter):
     """ The debug alerter uses a Python logger (by default, alerting to terminal). """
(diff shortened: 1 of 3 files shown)


elastalert/alerts.py
```diff
@@ -220,5 +210,36 @@ def alert(self, matches):
...
+            return self.create_custom_title(matches)
+
+        return self.create_default_title(matches)
+
+    def create_custom_title(self, matches):
+        alert_subject = self.rule['alert_subject']
+
+        if 'alert_subject_args' in self.rule:
+            alert_subject_args = self.rule['alert_subject_args']
+            alert_subject_values = [matches[0][arg] for arg in alert_subject_args]
```
Qmando: matches[0].get(arg)

elastalert/alerts.py
```diff
@@ -220,5 +210,36 @@ def alert(self, matches):
...
+            return alert_subject.format(*alert_subject_values)
```
Qmando: Can we get error handling around this?

jjsendor:
I have added the documentation as well as moved the `create_title` and `create_custom_title` to the `Alerter` class.
Also accommodated the comment about `get()` from CR.

elastalert/alerts.py
jjsendor: I have added now default value for get. Not sure what else it would make sense at this point.
Ensuring that the `alert_subjects_args` contains the exact number of elements to satisfy all `{#}` fields in `alert_subject`?

albarki:
sorry, this point need some explanation, 
alert_subject_args:
issue.name >> this is not return any thing. 
also what is the mean of "There are several ways to format the body text of the various types of events. In EBNF:" in this [docs](http://elastalert.readthedocs.org/en/latest/ruletypes.html#alert-content)

Qmando:
If you are getting "<MISSING VALUE>", then the match does not include that field. If you are not on the latest version, you may also need to add ``include: ['*']`` to your rule. You must also set ``alert_subject`` for ``alert_subject_args`` to be used.

EBNF is https://en.wikipedia.org/wiki/Extended_Backus%E2%80%93Naur_Form. The first secton shows the different components of the alert text, and then which components appear for each value of ``alert_text_type``. By default, it will look like:
```
Rule name

"At least X events matched at ...." or "There was an abnormal number of events ..." or something

@timestamp: 2015-08-26
issue.name: blahblah
```

If you set ``alert_text``, it will appear between "Rule name" and "At least X ...". If you add ``alert_text_type: alert_text_only``, it will only be ``alert_text``. If you set ``alert_text_type: exclude_fields``, it will remove the "@timestamp: 2015..." part. 

All of the alert types will have the same text, so if you set ``alert: debug`` or use ``--debug``, it will print the same text for an alert as it would email.

albarki:
@Qmando thanks, it works after the  update and adding `include: ['*']`
I like `elastalert`

PR #21, opened 2015-03-11
Top count query

Qmando:
Change top_count_keys from iterating through actual data to making an aggregation query to Elasticsearch. 

This is so top_count_keys can be used with use_count_query and use_terms_query. This was the major blocker in converting slow rules to use count instead of search+fetch. 

The function has been moved into the ElastAlerter class and is run by in it's alert function. I also changed the text "The following are the top X events by Y:" to just "Y:". 

I fixed the existing unit test and tested manually with frequency type rule both with and without query_key and with use_count_query and use_terms_query.


y-trobinso:
Ship It!

PR #23, opened 2015-03-12
Fix tox.ini to make travis work

Qmando:
Added py26 and py27 to tox envlist

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -4,9 +4,6 @@ envlist = py26,py27
 
 [testenv]
 deps = -rrequirements-dev.txt
-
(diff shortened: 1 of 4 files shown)


tox.ini
```diff
@@ -1,11 +1,12 @@
...
-envlist = py
+project = elastalert
+envlist = py26,py27
 
 [testenv]
 deps = -rrequirements-dev.txt
 
-[testenv:py]
-deps = {[testenv]deps}
+#[testenv:py26]
```
ivanlei: Delete these and then merge away.

PR #24, opened 2015-03-13
Convert timestamps

Qmando:
Convert timestamps to datetime objects in get_hits and convert them back in add_match. This makes ElastAlert MUCH faster, as a ton of time was spent converting back and forth while doing comparisons, especially when event_window and query_key were used together.

All tests were fixed to use the correct object type, ts_delta function was removed completely, alerts remain unchanged.

### DIFF
--- a/docs/source/recipes/adding_rules.rst
+++ b/docs/source/recipes/adding_rules.rst
@@ -34,8 +34,8 @@ option, this will be automatically converted to a ``datetime.timedelta`` object 
 ``self.matches``: This is where ElastAlert checks for matches from the rule. Whatever information is relevant to the match
 (generally comming from the fields in Elasticsearch) should be put into a dictionary object and
 added to ``self.matches``. ElastAlert will pop items out periodically and send alerts based on these objects. It is
-recommended that you use ``self.add_match(match)`` to add matches. By default, this simply appends to ``self.matches``,
-but some other rule types use it to add additional information to the match.
+recommended that you use ``self.add_match(match)`` to add matches. In addition to appending to ``self.matches``,
+``self.add_match`` will convert the datetime ``@timestamp`` back into an ISO 8601 timestamp.
 
 ``self.required_options``: This is a set of options that must exist in the configuration file. ElastAlert will
 ensure that all of these fields exist before trying to instantiate a ``RuleType`` instance.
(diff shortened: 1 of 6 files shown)


docs/source/recipes/adding_rules.rst
```diff
@@ -102,14 +104,15 @@ Now, in a file named ``my_rules.py``, add
...
+                    # Convert the timestamp to a time object
+                    login_time = document['@timestamp'].time()
 
+                    # Convert time_start and time_end to time objects
+                    time_start = dateutil.parser.parse(self.rule['time_start']).time()
+                    time_end = dateutil.parser.parse(self.rule['time_end']).time()
+                    
                     # If the time falls between start and end
-                    if str_time > self.rule['time_start'] and str_time < self.rule['time_end']:
+                    if str_time > time_start and str_time < time_end:
```
y-trobinso: This is fine, but be cautious if you ever deal with representations of unixtime serialized through these time fields. If you do, you'll be a second off, due to not setting these as >= and <=.

elastalert/elastalert.py
```diff
@@ -243,7 +248,7 @@ def remove_old_events(self, rule):
         remove = []
         buffer_time = rule.get('buffer_time', self.buffer_time)
         for _id, timestamp in rule['processed_hits'].iteritems():
-            if ts_delta(timestamp, now) > buffer_time:
+            if now - timestamp > buffer_time:
```
y-trobinso: Same comment as before about equivalence relations and Unixtime. Negligible.

tests/base_test.py
```diff
@@ -315,20 +318,19 @@ def test_count(ea):
...
+        query['query']['filtered']['filter']['bool']['must'][0]['range']['@timestamp']['to'] = dt_to_ts(end)
+        query['query']['filtered']['filter']['bool']['must'][0]['range']['@timestamp']['from'] = dt_to_ts(start)
+        start = start + ea.run_every
         ea.current_es.count.assert_any_call(body=query, doc_type='doctype', index='idx')
 
 
 def test_queries_with_rule_buffertime(ea):
     ea.rules[0]['buffer_time'] = datetime.timedelta(minutes=53)
-    hits = generate_hits([START])
+    hits = generate_hits([START_TS])
```
y-trobinso: This is very easy to confuse with "START_TLS". Consider a rename.

y-trobinso:
This change seems fine. I feel we can summarize some of these calls better into classes, but given the current code, what's here seems correct to me.

y-trobinso:
Verify the travis issue isn't a problem and ship it!

PR #26, opened 2015-03-17
Removed tox env from precommit hooks install

Qmando:
The pre-commit section in tox was removed in https://github.com/Yelp/elastalert/commit/d9ab6864d56908e9379ae9e0911a56432779e6da. Update the Makefile accordingly


jjsendor:
Ship it!

PR #29, opened 2015-03-18
Top count bug fix

Qmando:
Couple of fixes/features here:

1. Match timestamp fields are properly converted before calling get_top_counts
2. Query key is filtered by '.raw' field. With an analyzed field, passing the data to a term filter can result in an unexpected lack of hits.
3. Similarly, top_count_keys are defaulted to '.raw'. This is because otherwise they often come up as
"this": 5
"field": 5
"is": 5
"analyzed": 5

instead of

"this field is analyzed": 5

### DIFF
--- a/elastalert/config.py
+++ b/elastalert/config.py
@@ -157,6 +157,10 @@ def load_configuration(filename):
     include.append(rule['timestamp_field'])
     rule['include'] = list(set(include))
 
+    # Change top_count_keys to .raw
+    if 'top_count_keys' in rule and rule.get('raw_count_keys', True):
+	rule['top_count_keys'] = [key+'.raw' if not key.endswith('.raw') else key for key in rule.get('top_count_keys')] 
+
     # Check that generate_kibana_url is compatible with the filters
     if rule.get('generate_kibana_link'):
         for es_filter in rule.get('filter'):
--- a/elastalert/config.py
+++ b/elastalert/config.py
@@ -159,7 +159,8 @@ def load_configuration(filename):
 
     # Change top_count_keys to .raw
     if 'top_count_keys' in rule and rule.get('raw_count_keys', True):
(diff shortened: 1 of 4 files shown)


elastalert/config.py
```diff
@@ -157,6 +157,10 @@ def load_configuration(filename):
     include.append(rule['timestamp_field'])
     rule['include'] = list(set(include))
 
+    # Change top_count_keys to .raw
+    if 'top_count_keys' in rule and rule.get('raw_count_keys', True):
+        rule['top_count_keys'] = [key + '.raw' if not key.endswith('.raw') else key for key in rule.get('top_count_keys')]
```
y-trobinso: Long line. Otherwise seems fine.

elastalert/elastalert.py
```diff
@@ -675,8 +682,8 @@ def alert(self, matches, rule, alert_time=None):
                     qk = match[rule['query_key']]
                 else:
                     qk = None
-                start = ts_add(match[rule['timestamp_field']], -rule.get('timeframe', datetime.timedelta(minutes=10)))
-                end = ts_add(match[rule['timestamp_field']], datetime.timedelta(minutes=10))
+                start = ts_to_dt(match[rule['timestamp_field']]) - rule.get('timeframe', datetime.timedelta(minutes=10))
+                end = ts_to_dt(match[rule['timestamp_field']]) + datetime.timedelta(minutes=10)
```
y-trobinso: Extremely pedantic: at DST cutovers, you have a race condition that might widen or shorten the window if DST switches between when start and end are calculated.

It's unlikely to ever appear, but you should consider storing a current time variable to prevent this in future code.

y-trobinso:
Barring stylistic issues (long lines, implicit Unicode conversions) and the very outside chance of timespec issues around DST, this seems fine.

Ship It!

elastalert/elastalert.py
Qmando: I'll open up an issue for this. Low priority

PR #33, opened 2015-03-20
test_rule uses strf indexes and doesn't load alerts or rules

### LINKED ISSUE #20: elastalert-test-rule fails with use_strftime_index
The test rule script uses it's own code to query elasticsearch and it doesn't format the index if use_strftime_index is set.

Qmando:
Fixes #20 

Moved config stuff around so that we can do some checks and then return before loading any alerts/rules/enhancements.

### DIFF
--- a/elastalert/config.py
+++ b/elastalert/config.py
@@ -49,7 +49,7 @@ def get_module(module_name):
     return module
 
 
-def load_configuration(filename):
+def load_configuration(filename, testing=False):
     """ Load a yaml rule file and fill in the relevant fields with objects.
 
     :param filename: The name of a rule configuration file.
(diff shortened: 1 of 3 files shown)


elastalert/config.py
```diff
@@ -49,7 +49,7 @@ def get_module(module_name):
     return module
 
 
-def load_configuration(filename):
+def load_configuration(filename, testing=False):
```
ivanlei: Please add doc comments explaining what the `testing` flag is and how/when to use it.

In general, I think it's confusing to refer to this as `testing` since it's not unit testing.
Maybe `dry_run` or something?

Issue #8, opened 2015-03-02
Allow exponential backoff for realert


Qmando:
Instead of just having a fixed time before alerting again, there should be an option for exponential backoff.

PR #38, opened 2015-03-26
exponential_realert option

Qmando:
Silences now contain an exponent. By setting exponential_realert, realert will continue to double, up to exponential_realert, while the alerts are occurring within 2*realert of each other. If the time between alerts is longer, the exponent will decrease.

Fixes issue #8

### DIFF
--- a/elastalert/elastalert.py
+++ b/elastalert/elastalert.py
@@ -20,6 +20,7 @@ from util import dt_to_ts
 from util import EAException
 from util import format_index
 from util import pretty_ts
+from util import seconds
 from util import ts_add
 from util import ts_now
 from util import ts_to_dt
@@ -416,7 +417,8 @@ class ElastAlerter():
                 continue
 
             if rule['realert']:
-                self.set_realert(rule['name'] + key, dt_to_ts(datetime.datetime.utcnow() + rule['realert']))
+                next_alert, exponent = self.next_alert_time(rule, rule['name'] + key, ts_now())
+                self.set_realert(rule['name'] + key, next_alert, exponent)
 
             # If no aggregation, alert immediately
             if not rule['aggregation']:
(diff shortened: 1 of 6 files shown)


elastalert/elastalert.py
```diff
@@ -978,6 +982,33 @@ def get_top_counts(self, rule, starttime, endtime, keys, number=5, qk=None):
             all_counts['top_events_%s' % (key)] = dict(counts[:number])
         return all_counts
 
+    def next_alert_time(self, rule, name, timestamp):
+        """ Calculate an 'until' time and exponent based on how much past the last 'until' we are. """
+        if name in self.silence_cache:
+            last_until, exponent = self.silence_cache[name]
+        else:
+            # If for some reason this isn't cached, pretend it was a week ago
```
y-trobinso: This seems arbitrary. Do you have date to back up whether this is reasonable?

tests/base_test.py
```diff
@@ -498,3 +498,31 @@ def test_count_keys(ea):
...
+
+    test_values = [(ts5s, until, 0),  # Exp will increase to 1, 10*2**0 = 10s
+                   (ts15s, until, 0),  # Exp will stay at 0, 10*2**0 = 10s
+                   (ts15s, until, 1),  # Exp will increase to 2, 10*2**1 = 20s
+                   (ts1m, until, 2),  # Exp will decrease to 1, 10*2**2 = 40s
+                   (ts1m, until, 3),  # Exp will increase to 4, 10*2**3 = 1m20s
+                   (ts5m, until, 1),  # Exp will lower back to 0, 10*2**1 = 20s
+                   (ts4h, until, 9),  # Exp will lower back to 0, 10*2**9 = 1h25m
+                   (ts4h, until, 10),  # Exp will lower back to 9, 10*2**10 = 2h50m
+                   (ts4h, until, 11)]  # Exp will increase to 12, 10*2**11 = 5h
```
y-trobinso: Might be worth horizontally aligning these comments. Not a huge concern.

elastalert/elastalert.py
```diff
@@ -978,6 +982,33 @@ def get_top_counts(self, rule, starttime, endtime, keys, number=5, qk=None):
...
+            last_until = timestamp - datetime.timedelta(weeks=1)
+            exponent = 0
+
+        if not rule.get('exponential_realert'):
+            return timestamp + rule['realert'], 0
+
+        diff = seconds(timestamp - last_until)
+        # If it's been less than twice the alert time, increase exponent
```
y-trobinso: In general, I prefer comments that explain why instead of how. Perhaps prose explaining why this works would be more effective here? Otherwise, it's better to remove these comments, as they're not doing anything we can't already read here.

elastalert/elastalert.py
```diff
@@ -978,6 +982,33 @@ def get_top_counts(self, rule, starttime, endtime, keys, number=5, qk=None):
...
+        if diff < seconds(rule['realert']) * 2 ** exponent:
+            exponent += 1
+        else:
+            # If it's been longer, recursively decrease exponent and compare the remaining time difference
```
y-trobinso: Nit: this is iterative, not recursive. :)

elastalert/elastalert.py
```diff
@@ -978,6 +982,33 @@ def get_top_counts(self, rule, starttime, endtime, keys, number=5, qk=None):
...
+            while diff > seconds(rule['realert']) * 2 ** exponent and exponent > 0:
```
y-trobinso: This is acceptable, but I believe you're doing a projection (http://en.wikipedia.org/wiki/Projection_%28linear_algebra%29) iteratively here. You can knock this down from O(n) to O(1).

In practice, the diff will be relatively small, so this won't consume many cycles. But it's worth mentioning here.

y-trobinso:
This is fine to merge. I've posted my nits above.

Issue #35, opened 2015-03-25
Add support for basic-auth and SSL


bitsofinfo:
Would be nice to be able to have support for SSL and basic-auth

Qmando:
Support for basic-auth was added for writeback_es (metadata) in https://github.com/Yelp/elastalert/pull/36. Support still needs to be added for current_es (queries).

bitsofinfo:
I'll take care of this

PR #39, opened 2015-03-26
ssl/auth es construction change for #35

bitsofinfo:
ssl/auth es construction change for #35


Qmando:
This is great! Thanks for adding this.

Issue #40, opened 2015-03-26
SpikeRule will alert again if alert_on_new_data is set and threshold_ref isn't


Qmando:
If alert_on_new_data is set, elastalert will skip the check that the reference window has been filled before sending an alert. This is appropriate if a unique term hasn't been seen before, but not if an alert was just triggered.

PR #41, opened 2015-03-27
Added another spike skip condition for after an alert

Qmando:
Fixes a bug where if alert_on_new_data is set, after an alert, checking for alert conditions will resume immediately, triggering an unexpected alert because ref_window was cleared. Previous, there were two conditions for skipping checks:

1. Not enough time passed for this qk and alert_on_new_data is not set
2. Not enough time passed for any qk

This adds a third condition

3. Not enough time passed for this qk and not enough time has passed since an alert for this key

Fixes issue #40


y-trobinso:
This is a workaround, but after some discussion off thread, it seems fine to me. Ship It!

PR #50, opened 2015-04-02
Changed example to actually work

Qmando:
1. example_change was missing a required ignore_null field.
2. The invocation of elastalert as a file causes the module to attempt import from the wrong location.
3. There were two errors in the example alerter.
4. Import errors were cryptic


y-trobinso:
Ship It!

PR #56, opened 2015-04-08
Added unhashable compatibility for ChangeRule query_key

Qmando:
If the query_key was a list for example, ChangeRule would fail when add_match attempts to access the key in the match.


y-trobinso:
lgtm

PR #57, opened 2015-04-09
Unhashable qk

Qmando:
Fixed error when trying to construct "rulename.query_key" when query_key is a list.
Added support for lists in kibana add_filter.
Fixed datetime's not being converted if timestamp_field wasn't the default.
Fixed type=any rule skipping timestamp conversion.
Added a test for kibana.
Fixed a test that broke when I fixed the timestamp conversion. (Hits was reused, broke when timestamp field was converted in add_match)

### DIFF
--- a/elastalert/kibana.py
+++ b/elastalert/kibana.py
@@ -200,7 +200,12 @@ def add_filter(dashboard, es_filter):
         if isinstance(f_query, basestring):
             f_query = '"%s"' % (f_query.replace('"', '\\"'))
         if isinstance(f_query, list):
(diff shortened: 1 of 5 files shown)


elastalert/kibana.py
```diff
@@ -199,6 +199,8 @@ def add_filter(dashboard, es_filter):
         # Wrap query in quotes, otherwise certain characters cause kibana to throw errors
         if isinstance(f_query, basestring):
             f_query = '"%s"' % (f_query.replace('"', '\\"'))
+        if isinstance(f_query, list):
+            f_query = '(%s)' % (' AND '.join(['"%s"' % (item.replace('"', '\\"')) for item in f_query]))
```
ivanlei: This code is too dense. Please break into multiple lines.

PR #64, opened 2015-04-15
Added scan_subdirectories option to load rules recursively

Qmando:
This adds the option scan_subdirectories, which allows you to recursively load rules from rules_folder and it's subdirectories. I added a basic unit test for getting file paths. I also manually tested this and verified that it correctly picked up on changes.

### DIFF
--- a/docs/source/elastalert.rst
+++ b/docs/source/elastalert.rst
@@ -108,11 +108,8 @@ even if ElastAlert is stopped and restarted, it will never miss data or look at 
 ``es_password``: Optional; basic-auth password for connecting to ``es_host``.
 
 ``rules_folder``: The name of the folder which contains rule configuration files. ElastAlert will load all
-files in this folder that end in .yaml. If the contents of this folder change, ElastAlert will load, reload
+files in this folder, and all subdirectories, that end in .yaml. If the contents of this folder change, ElastAlert will load, reload
 or remove rules based on their respective config files.
-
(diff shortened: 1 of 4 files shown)


docs/source/elastalert.rst
```diff
@@ -111,6 +111,9 @@ even if ElastAlert is stopped and restarted, it will never miss data or look at
 files in this folder that end in .yaml. If the contents of this folder change, ElastAlert will load, reload
 or remove rules based on their respective config files.
 
+``scan_subdirectories``: If true, ElastAlert will recursively scan ``rules_folder`` for yaml files to load as rules.
```
ivanlei: I feel like `scan_subdirectories: true` should just be the behavior and we could remove the option.
Sometimes there's a LOT of options.

elastalert/config.py
```diff
@@ -238,15 +250,14 @@ def load_rules(filename, use_rule=None):
         raise EAException('Invalid time format used: %s' % (e))
 
     # Load each rule configuration file
-    rule_folder = conf['rules_folder']
-    rule_files = os.listdir(rule_folder)
     rules = []
+    rule_files = get_file_paths(conf)
     for rule_file in rule_files:
         if use_rule and rule_file != use_rule:
```
talyian: use_rule now needs the full path -- could you lump the use_rule and file extension check into get_file_paths to reduce duplication?

talyian:
🚢

PR #65, opened 2015-04-16
Allow callers to run rules using their own looping construct, or to stop the elastalert loop themselves

jaredbranum:
This should give callers greater control over starting and stopping the elastalert rule running loop. You could either call `run_all_rules` directly within your own loop, or call `stop` whenever you're ready to exit the elastalert loop. The latter approach is a bit trickier for the caller, since `start` won't return until `stop` is called, so you'll need to manage separate threads (or something similar).

### DIFF
--- a/elastalert/elastalert.py
+++ b/elastalert/elastalert.py
@@ -560,7 +560,8 @@ class ElastAlerter():
             except (TypeError, ValueError):
                 self.handle_error("%s is not a valid ISO 8601 timestamp (YYYY-MM-DDTHH:MM:SS+XX:00)" % (self.starttime))
                 exit(1)
-        while True:
+        self.running = True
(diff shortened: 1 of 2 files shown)


elastalert/elastalert.py
```diff
@@ -553,62 +554,70 @@ def load_rule_changes(self):
...
+        self.starttime = None
 
-            # Only force starttime once
-            starttime = None
+        if not self.args.pin_rules:
+            self.load_rule_changes()
 
-            if not self.args.pin_rules:
-                self.load_rule_changes()
+        # Wait before querying again
```
Qmando: I think this should go in the start loop. Sleeping before returning makes no sense if you are calling run_all_rules directly.

PR #67, opened 2015-04-16
Calling garbage collect after rule runs; test cleanup.

eastebry:
Doing a few things
* garbage collect every time a rule runs - should fix  issue with flatline rule not properly alerting when run over historical data
* add tests
* move replace_ts to util, since it doesn't really need to be a class function
* Fix all the warnings cause by `ts_to_dt` an `dt_to_ts` in test code
* Small cleanup

### DIFF
--- a/elastalert/elastalert.py
+++ b/elastalert/elastalert.py
@@ -416,14 +416,13 @@ class ElastAlerter():
         # Run the rule
         # If querying over a large time period, split it up into chunks
         self.num_hits = 0
-        tmp_endtime = endtime
         buffer_time = rule.get('buffer_time', self.buffer_time)
         while endtime - rule['starttime'] > buffer_time:
             tmp_endtime = rule['starttime'] + self.run_every
             if not self.run_query(rule, rule['starttime'], tmp_endtime):
                 return 0
             rule['starttime'] = tmp_endtime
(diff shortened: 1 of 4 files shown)


elastalert/elastalert.py
```diff
@@ -426,11 +423,10 @@ def run_rule(self, rule, endtime, starttime=None):
             if not self.run_query(rule, rule['starttime'], tmp_endtime):
                 return 0
             rule['starttime'] = tmp_endtime
+            rule['type'].garbage_collect(endtime)
         if not self.run_query(rule, rule['starttime'], endtime):
             return 0
 
-        rule['type'].garbage_collect(endtime)
```
Qmando: This shouldn't be removed. If the loop is never entered, run_query will still be called and garbage_collect still needs to be run.

elastalert/elastalert.py
```diff
@@ -967,6 +967,10 @@ def set_realert(self, rule_name, timestamp, exponent):
 
     def is_silenced(self, rule_name):
         """ Checks if rule_name is currently silenced. Returns false on exception. """
+
+        if self.debug:
```
Qmando: I think this should go below the check for silence_cache. Debug should still have normal realert behavior, minus the fetching from Elasticsearch part.

PR #69, opened 2015-04-17
Fixed race condition in test_stop

Qmando:
An assert came directly after thread.start(), not giving it time to create the ea.running variable.

Context is this change: https://github.com/Yelp/elastalert/pull/65.

Tests fail sometimes with an AttributeError and pass most of the time


Qmando:
Using a different approach.

PR #72, opened 2015-04-21
Added CommandAlerter for running arbitrary commands as an alert

Qmando:
An alerter for running arbitrary commands. The command arguments can be string formatted using the match. You are also able to pass the match as json to stdin. Theres currently no support for looking at process exit codes or stdout.

Note: This uses the % formatting instead of str.format because it allows you to use objects from a dictionary without specifying an index, ie %(fieldname)s vs {0[fieldname]}

### DIFF
--- a/docs/source/ruletypes.rst
+++ b/docs/source/ruletypes.rst
@@ -498,7 +498,7 @@ until the command exits or sends an EOF to stdout.
 
 Example usage::
 
-    alerts:
+    alert:
       - command
     command: ["/bin/send_alert", "--username", "%(username)s"]
 
--- a/docs/source/ruletypes.rst
+++ b/docs/source/ruletypes.rst
@@ -478,6 +478,36 @@ field_values will contain every key value pair included in the results from Elas
 every key in ``included``, every key in ``top_count_keys``, ``query_key``, and ``compare_key``. If the alert spans multiple events, these values may
 come from an individual event, usually the one which triggers the alert.
 
+Command
+~~~~~~~
+
+The command alert allows you to execute an arbitrary command and pass arguments or stdin from the match. Arguments to the command can use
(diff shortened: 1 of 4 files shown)


docs/source/ruletypes.rst
```diff
@@ -478,6 +478,36 @@ field_values will contain every key value pair included in the results from Elas
...
+match passed the format argument. This means that a field can be accessed with ``%(field_name)s``.
+
+Optional:
+
+``pipe_match_json``: If true, the match will be converted to JSON and passed to stdin of the command. Note that this will cause ElastAlert to block
+until the command exits or sends an EOF to stdout.
+
+Example usage::
+
+    alerts:
```
talyian: alerts or alert?

Qmando: Perhaps alerts would make more sense as it can take a list.. but I should probably use valid examples in the docs ;)

docs/source/ruletypes.rst
```diff
@@ -478,6 +478,36 @@ field_values will contain every key value pair included in the results from Elas
...
+      - command
+    command: ["/bin/send_alert", "--username", "%(username)s"]
```
talyian: multiple commands would  be nice

elastalert/alerts.py
```diff
@@ -292,3 +296,37 @@ def create_default_title(self, matches, for_search=False):
...
+    def __init__(self, *args):
+        super(CommandAlerter, self).__init__(*args)
+        if isinstance(self.rule['command'], basestring) and '%' in self.rule['command']:
+            logging.warning('Warning! You could be vulnerable to shell injection!')
+            self.rule['command'] = [self.rule['command']]
+
+    def alert(self, matches):
+        for match in matches:
+            # Format the command and arguments
+            command = copy.copy(self.rule['command'])
```
talyian: why the copy?

PR #76, opened 2015-04-23
Fix race condition

Qmando:
There was a race condition in test_stop in which ea.running raised an Exception because the thread had not started yet. Here we mock run_all_rules to verify it runs multiple times, give the thread a few seconds to exit, and fail if it does not.

I tested this by commenting out the code in ElastAlerter.stop and verified that the test failed and didn't cause pytest to hang

### DIFF
--- a/tests/base_test.py
+++ b/tests/base_test.py
@@ -4,7 +4,6 @@ import copy
 import datetime
 import json
 import threading
-import time
 
 import elasticsearch
 import mock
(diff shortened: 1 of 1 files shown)


talyian:
I'd use the iterator form of side_effect here. https://pb.yelpcorp.com/129662

Qmando:
Ah thank you! I knew using a parent scope list was a pretty hack way of keeping state in the side effect. That is much nicer

Issue #61, opened 2015-04-15
is_silenced causes error on new index


Qmando:
create_index only creates a mapping for the silence type for rule_name. When is_silenced tries to query it sorts by "until" which doesn't yet exist in the index. The mapping should be specified for every type.

gopalganu:
I am using version 0.0.44 and get the following error.
This is a frequency rule. What may be the reason.


elastalert_status/silence/_search?_source_include=until%2Cexponent&size=1 [status:400 request:0.005s]
ERROR:root:Error while querying for alert silence status: TransportError(400, u'SearchPhaseExecutionException[Failed to execute phase [query_fetch], all shards failed; shardFailures {[G4zGZ6nPSwaOPw_1KoF9Iw][elastalert_status][0]: SearchParseException[[elastalert_status][0]: from[-1],size[-1]: Parse Failure [Failed to parse source [{"filter": {"term": {"rule_name": "instrumentation-errors.56E77923-D99D-4962-88CC-04DDE460EB18"}}, "sort": {"until": {"order": "desc"}}}]]]; nested: SearchParseException[[elastalert_status][0]: from[-1],size[-1]: Parse Failure [No mapping found for [until] in order to sort on]]; }]')

Qmando:
@gopalganu Did run create_index.py? It should upload a mapping to the elastalert_status which will prevent this error.

gopalganu:
I didn't run it manually, do I have to?

PR #77, opened 2015-04-23
Added until to silence_mapping

Qmando:
Fixes issue #61 

Tested manually by creating a new index, pointing elastalert to the new index, ran a rule which triggered an alert, and set a breakpoint verifying elastalert made the query in is_silenced.


y-trobinso:
lgtm

Issue #83, opened 2015-05-06
terms_size is undocumented


Qmando:
terms_size, the maximum number of results returned when using use_terms_query, is undocumented and has a horrible default of just 5

PR #84, opened 2015-05-06
Added docs for terms_size, upped the default, and fixed top_count_number

### LINKED ISSUE #73: top_count_number sometimes doesn't work
get_top_counts will call get_terms_hits without passing a size limit, and so get_terms_hits will use "terms_size" or it's default of 5

Qmando:
This fixes #73 and #83. terms_size was undocumented and had a ridiculously low default. Also, it overrode top_count_number because the code was shared.


ivanlei:
@mlw and @Qmando let's complete the review and push this sucker.

Issue #87, opened 2015-05-13
Handle JIRA search errors


Qmando:
When creating a ticket, ElastAlert will search for an existing one, if something like traceback[10] is in the title, JIRA will throw an error.

We should gracefully handle these errors.

PR #89, opened 2015-05-13
Added exception handling for jira.search_issues

Qmando:
Fixes issue #87. If jira.client.search_issues throws an exception, which can happen if the summary (which contains the value of query_key) contains special characters, handle it as if no results were returned.

This also adds a test case for this scenario.


y-trobinso:
Ship It!

PR #90, opened 2015-05-15
Add new term rule

Qmando:
A new rule type which will search for a list of existing terms when it starts, and then alert on new values. It can take a number of different fields, and can use_terms_query.

### DIFF
--- a/elastalert/ruletypes.py
+++ b/elastalert/ruletypes.py
@@ -486,7 +486,7 @@ class NewTermsRule(RuleType):
             self.fields = self.rules['fields']
         if type(self.fields) != list:
             self.fields = [self.fields]
-        if self.rules.get('use_terms_query') and len(self.fields) > 1:
+        if self.rules.get('use_terms_query') and len(self.fields) != 1:
             raise EAException("use_terms_query can only be used with one field at a time")
         self.get_all_terms()
 
@@ -523,6 +523,18 @@ class NewTermsRule(RuleType):
                         document['new_field'] = field
                         self.add_match(document)
                         self.seen_values[field].append(value)
+
+    def add_terms_data(self, terms):
+        # With terms query, len(self.fields) is always 1
+        field = self.fields[0]
+        for timestamp, buckets in terms.iteritems():
+            for bucket in buckets:
+                if bucket['doc_count']:
+                    if bucket['key'] not in self.seen_values[field]:
+                        match = {field: bucket['key'],
+                                 self.rules['timestamp_field']: timestamp,
+                                 'new_field': field}
+                        self.add_match(match)
 
     def garbage_collect(self, ts):
         pass
--- a/elastalert/ruletypes.py
+++ b/elastalert/ruletypes.py
@@ -2,11 +2,14 @@
 import datetime
 from collections import deque
 
+from elasticsearch.client import Elasticsearch
 from util import dt_to_ts
 from util import EAException
+from util import format_index
 from util import hashable
 from util import lookup_es_key
 from util import pretty_ts
+from util import ts_now
 from util import ts_to_dt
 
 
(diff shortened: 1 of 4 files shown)


elastalert/ruletypes.py
```diff
@@ -466,3 +469,72 @@ def garbage_collect(self, ts):
...
+        if self.rules.get('use_terms_query') and len(self.fields) != 1:
+            raise EAException("use_terms_query can only be used with one field at a time")
+        self.get_all_terms()
+
+    def get_all_terms(self):
+        """ Performs a terms aggregation for each field to get every existing term. """
+        self.es = Elasticsearch(host=self.rules['es_host'], port=self.rules['es_port'])
+        window_size = datetime.timedelta(**self.rules.get('terms_window_size', {'days': 30}))
+
+        field_name = {"field": "", "size": 999999}
```
Qmando: The Elasticsearch docs say that by passing size: 0, its equivalent to size: MAXINT. In my testing however, passing 0 never returned any results. This is an arbitrary magic number.

elastalert/ruletypes.py
```diff
@@ -466,3 +469,72 @@ def garbage_collect(self, ts):
...
+        if self.rules.get('use_strftime_index'):
+            end = ts_now()
+            start = end - window_size
+            index = format_index(self.rules['index'], start, end)
+        else:
+            index = self.rules['index']
+
+        for field in self.fields:
+            field_name['field'] = field
+            res = self.es.search(body=query_template, index=index, ignore_unavailable=True, timeout=50)
```
Qmando: Timeout here is also a magic number. Since this is a really big query, it can easily take a long time. (Longer than the default 10s timeout at least)

elastalert/ruletypes.py
```diff
@@ -466,3 +469,72 @@ def garbage_collect(self, ts):
...
+        # Allow the use of query_key or fields
+        if 'fields' not in self.rules:
+            if 'query_key' not in self.rules:
+                raise EAException("fields or query_key must be specified")
+            self.fields = self.rules['query_key']
+        else:
+            self.fields = self.rules['fields']
+        if type(self.fields) != list:
+            self.fields = [self.fields]
+        if self.rules.get('use_terms_query') and len(self.fields) != 1:
```
y-trobinso: This will be confusing if self.fields ever has a length of zero.

elastalert/ruletypes.py
y-trobinso: This constant seems arbitrary. Can we name size, or use something astronomically large like MAX_INT, if it's the default?

y-trobinso: I think we should actually use python's version of maxint here, instead of an arbitrary number we might surpass.

elastalert/ruletypes.py
```diff
@@ -466,3 +469,72 @@ def garbage_collect(self, ts):
...
+            end = ts_now()
+            start = end - window_size
+            index = format_index(self.rules['index'], start, end)
+        else:
+            index = self.rules['index']
+
+        for field in self.fields:
+            field_name['field'] = field
+            res = self.es.search(body=query_template, index=index, ignore_unavailable=True, timeout=50)
+            buckets = res['aggregations']['values']['buckets']
```
y-trobinso: I really wish we didn't have to do nested traversal like this, since it violates the principle of least knowledge. Outside of turning these into classes, however, I don't have cogent advice on a cleanup.

elastalert/ruletypes.py
```diff
@@ -466,3 +469,72 @@ def garbage_collect(self, ts):
...
+        for timestamp, buckets in terms.iteritems():
+            for bucket in buckets:
+                if bucket['doc_count']:
+                    if bucket['key'] not in self.seen_values[field]:
+                        match = {field: bucket['key'],
+                                 self.rules['timestamp_field']: timestamp,
+                                 'new_field': field}
+                        self.add_match(match)
+
+    def garbage_collect(self, ts):
```
y-trobinso: Is this simply unimplemented? Where's the TODO?

Qmando: Ah oops. I was thinking in my head the base class had NotImplementedError, but it's actually just a pass as well.

elastalert/ruletypes.py
Qmando: Elasticsearch is the epitome of nested traversal. It's unavoidable, other than doing it one line at a time.

y-trobinso:
Ship it!

Issue #68, opened 2015-04-17
Pipeline alerts and allow them to share data


Qmando:
The alerts should have some way of communicating with each other. This could be done by creating a pipeline in the order defined in the rule file, where each alerter can pass a dictionary of data to the next. This would allow, for example, a JIRA alert to create a ticket and then an email alert to contain a link to that ticket.

PR #91, opened 2015-05-19
Alert pipeline

Qmando:
Fixes issue #68 

Each alerter is given a shared dictionary object which can be used to transfer information between alerters. This allows the rules with both JIRA and email alerters to have the ticket created linked in the email. The order in which the alerters are defined matters in this case. I'm not sure whether this warrants a check in config to automatically reorder email to always come after JIRA if they both exist.

### DIFF
--- a/docs/source/recipes/adding_alerts.rst
+++ b/docs/source/recipes/adding_alerts.rst
@@ -34,7 +34,7 @@ should be in the rule configuration file and can be accessed here.
 
 ``self.pipeline``: This is a dictionary object that serves to transfer information between alerts. When an alert is triggered,
 a new empty pipeline object will be created and each alerter can add or receive information from it. Note that alerters
-are called in the order they are defined in the rule file. For example, the JIRA alerter will add it's ticket number
+are called in the order they are defined in the rule file. For example, the JIRA alerter will add its ticket number
 to the pipeline and the email alerter will add that link if it's present in the pipeline.
 
 alert(self, match):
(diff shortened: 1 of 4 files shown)


docs/source/recipes/adding_alerts.rst
```diff
@@ -32,6 +32,11 @@ present. ElastAlert will not instantiate the alert if any are missing.
 ``self.rule``: The dictionary containing the rule configuration. All options specific to the alert
 should be in the rule configuration file and can be accessed here.
 
+``self.pipeline``: This is a dictionary object that serves to transfer information between alerts. When an alert is triggered,
+a new empty pipeline object will be created and each alerter can add or receive information from it. Note that alerters
+are called in the order they are defined in the rule file. For example, the JIRA alerter will add it's ticket number
```
y-trobinso: Spelling: "its"

elastalert/elastalert.py
```diff
@@ -776,7 +776,9 @@ def alert(self, matches, rule, alert_time=None):
         # Run the alerts
         alert_sent = False
         alert_exception = None
+        alert_pipeline = {}
         for alert in rule['alert']:
+            alert.pipeline = alert_pipeline
```
y-trobinso: alert_pipeline appears to always be an empty dictionary. Either turn this into a proper constant (ie, just name it ALERT_PIPELINE, to signify its place as a constant), or eliminate it entirely and set alert.pipeline to {} on every run.

An extra caveat here is alert_pipeline is a pointer to the same dictionary. This may cause some surprising behavior if the caller does not use a deep copy: you'll end up with the exact same pointer propagated to every rule you cut in your loop! I suspect that is not what you desire here.

As such, I vote for eliminating the extra variable to remove this surprising behavior, or if you are relying upon it, adding a comment explicitly calling out the pointer reuse.

y-trobinso:
Fix-and-ship ^

elastalert/elastalert.py
Qmando: Sharing the same object is the the point of this, so that the first alert can put information into it and the second can use that information. I'll add a comment.

PR #92, opened 2015-05-20
Added from_addr option to email alerter

eastebry:
Added a from_addr option that allows you to specify the email from address in the rule.


Qmando:
Looks good to me! Please add a section in the docs for from_addr.

PR #96, opened 2015-05-22
Added initial support for kibana4 links

Qmando:
This adds the option "use_kibana4_dashboard", which when used with "kibana_url", will put a link in the alert to 
<kibana_url>/#/dashboard/<use_kibana4_dashboard>?_g=<time range>

I tried getting filters to work, but they are slightly more complicated. That can come later.

Tested with Kibana 4.0.2

### DIFF
--- a/docs/source/ruletypes.rst
+++ b/docs/source/ruletypes.rst
@@ -196,9 +196,9 @@ ElastAlert can use an existing dashboard. It will set the time range on the dash
 upload it as a temporary dashboard, add a filter to the ``query_key`` of the alert if applicable,
 and put the url to the dashboard in the alert. (Optional, string, no default)
 
-``use_kibana4_dashboard``: The name of a Kibana 4 dashboard to link to. This will set the time setting on the dashboard
-from the match time minus the timeframe, to 10 minutes after the match time. Note that this does not support filtering
-by ``query_key`` like Kibana 3.
+``use_kibana4_dashboard``: A link to a Kibana 4 dashboard. For example, "https://kibana.example.com/#/dashboard/My-Dashboard". 
+This will set the time setting on the dashboard from the match time minus the timeframe, to 10 minutes after the match time. 
+Note that this does not support filtering by ``query_key`` like Kibana 3.
 
 ``use_local_time``: Whether to convert timestamps to the local time zone in alerts. If false, timestamps will
 be converted to UTC, which is what ElastAlert uses internally. (Optional, boolean, default true)
--- a/docs/source/ruletypes.rst
+++ b/docs/source/ruletypes.rst
@@ -188,7 +188,7 @@ dashboard will also contain a filter for the ``query_key`` of the alert. The das
 be uploaded to the kibana-int index as a temporary dashboard. (Optional, boolean, default False)
 
 ``kibana_url``: The url to access Kibana. This will be used if ``generate_kibana_link`` or
-``use_kibana[4]_dashboard`` is true. This must be specified with ``use_kibana4_dashboard``.
+``use_kibana_dashboard`` is true. If not specified, a URL will be constructed using ``es_host`` and ``es_port``.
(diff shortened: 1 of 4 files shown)


elastalert/config.py
```diff
@@ -151,6 +151,10 @@ def load_options(rule):
                                 'The index will be formatted like %s' % (token,
                                                                          datetime.datetime.now().strftime(rule.get('index'))))
 
+    # Kibana4 urls can't be generated based on host:port
```
striglia: Or more precisely, Kibana4 host+port cannot be assumed to be the same as the ES index ElastAlert is querying.

docs/source/ruletypes.rst
```diff
@@ -181,19 +181,25 @@ field name plus ".raw" to count unanalyzed terms. To turn this off, set ``raw_co
...
+``use_kibana[4]_dashboard`` is true. This must be specified with ``use_kibana4_dashboard``.
 (Optional, string, default ``http://<es_host>:<es_port>/_plugin/kibana/``)
 
-``use_kibana_dashboard``: The name of a dashboard to link to. Instead of generating a dashboard from a template,
+``use_kibana_dashboard``: The name of a Kibana 3 dashboard to link to. Instead of generating a dashboard from a template,
 ElastAlert can use an existing dashboard. It will set the time range on the dashboard to around the match time,
 upload it as a temporary dashboard, add a filter to the ``query_key`` of the alert if applicable,
 and put the url to the dashboard in the alert. (Optional, string, no default)
 
+``use_kibana4_dashboard``: The name of a Kibana 4 dashboard to link to. This will set the time setting on the dashboard
```
striglia: Is it better to just include the full URL here? That would give users the flexibility to specify whatever they want (search/viz/dash) and eliminate the extra kibana_url config field. These all should support explicit search queries (actually need to check that for dashes w/ saved searches) and custom time spans.

elastalert/kibana.py
```diff
@@ -166,6 +166,8 @@
                   u'style': u'dark',
                   u'title': u'ElastAlert Alert Dashboard'}
 
+kibana4_time_temp = "(refreshInterval:(display:Off,section:0,value:0),time:(from:'%s',mode:absolute,to:'%s'))"
```
striglia: Can we either set a non-zero refresh time or allow this to be customized? Would be pretty easy to have this templated by field name, then splat in a dict straight from the YAML config of whatever vars the user wanted to add.

Too loosely typed maybe?

docs/source/ruletypes.rst
Qmando: This is a good idea. With Kibana 3 you only needed the dashboard name, but with 4 you have to specify both the URL and dashboard name.

elastalert/kibana.py
Qmando: I think it should default to no autorefresh. I personally find it obnoxious sometimes, and with a set time window, these dashboards are likely to not change after an alert fires unless data is being backfilled. Having the option would be nice though.

striglia: Good point on the set time window. Given that, I think punting auto-refresh makes sense. If we were to ever support more flexible versions of the time windowing, allowing auto-refresh might make more sense.

docs/source/ruletypes.rst
striglia: Yeah...I guess the idea is with K3 there was no configuration you could do in URL anyways :) Now we have all the parameters available if we want them, so it may make sense for ElastAlert to allow more configuration where possible.

elastalert/kibana.py
y-trobinso: I also have a style concern here: please include spaces after commas for legibility and PEP8.

Qmando: I don't think Kibana4 will work if there are spaces and I don't think it's worth it to add spaces only to strip them out immediately.

talyian:
lg2m

Issue #95, opened 2015-05-21
Don't allow email_reply_to to be a list


Qmando:
If email_reply_to is a list, smtplib will throw an exception which doesn't get caught

Qmando:
Closing this as fixed by #102

PR #99, opened 2015-05-28
Siem 2578/bfrisbie/issue95

bfrizb:
Adds a catch for any Exceptions thrown by alerters.

This is in general to prevent ElastAlert from crashing on unexpected exceptions thrown by alerters.

In particular it's to address https://github.com/Yelp/elastalert/issues/95, where an smtplib error (from a buggy rule) can crash ElastAlert.

I ran "make test" and all tests passed.

### DIFF
--- a/elastalert/elastalert.py
+++ b/elastalert/elastalert.py
@@ -784,7 +784,10 @@ class ElastAlerter():
             try:
                 alert.alert(matches)
             except EAException as e:
-                self.handle_error('Error while running alert %s: %s' % (alert.get_info()['type'], e), {'rule': rule['name']})
+                self.handle_error('ElastAlert Error while running alert %s: %s' % (alert.get_info()['type'], e), {'rule': rule['name']})
+                alert_exception = str(e)
+            except Exception as e:
+                self.handle_error('Unexpected Error while running alert %s: %s' % (alert.get_info()['type'], e), {'rule': rule['name']})
                 alert_exception = str(e)
             else:
                 self.alerts_sent += 1
--- a/setup.py
+++ b/setup.py
@@ -8,7 +8,7 @@ from setuptools import setup
 base_dir = os.path.dirname(__file__)
 setup(
     name='elastalert',
-    version='0.0.38',
+    version='0.0.39',
     description='Runs custom filters on Elasticsearch and alerts on matches',
     author='Quentin Long',
     author_email='qlo@yelp.com',
--- a/tests/base_test.py
+++ b/tests/base_test.py
@@ -10,6 +10,8 @@ import mock
 import pytest
 from elasticsearch.exceptions import ElasticsearchException
 
+from alerts_test import mock_rule
+from elastalert.alerts import EmailAlerter
 from elastalert.enhancements import BaseEnhancement
 from elastalert.kibana import dashboard_temp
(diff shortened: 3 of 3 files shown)


elastalert/elastalert.py
```diff
@@ -784,7 +784,10 @@ def alert(self, matches, rule, alert_time=None):
             try:
                 alert.alert(matches)
             except EAException as e:
-                self.handle_error('Error while running alert %s: %s' % (alert.get_info()['type'], e), {'rule': rule['name']})
+                self.handle_error('ElastAlert Error while running alert %s: %s' % (alert.get_info()['type'], e), {'rule': rule['name']})
+                alert_exception = str(e)
+            except Exception as e:
+                self.handle_error('Unexpected Error while running alert %s: %s' % (alert.get_info()['type'], e), {'rule': rule['name']})
```
y-trobinso: You'll want to be cautious here. The class `Exception` includes `SyntaxError` and `RuntimeException`, which can make crafting new rules and identifying their failure scenarios problematic. This is a form of [Pokemon Exception Handling](http://c2.com/cgi/wiki?PokemonExceptionHandling) that may have unexpected ramifications in developing and testing alerts. This also obliterates the traceback for the error, further confounding development.

In practice, this is exactly why we created EAException: it allowed exception handling to be pushed down into the rules themselves  instead of relying on the scheduler.

Errors that crash ElastAlert itself should be as loud as possible, but I agree that they should not crash the application. An alternative to this approach would be to provide a thread pool for rules, reporting and reclaiming any crashed threads along the way. This can be done with a thread headcount of 1 to start with until single-threaded assumptions are resolved. Threading is something we want to do anyway, which further strengthens this recommendation.

In short: let's not make it easy for rules authors to write flaky rules that rely on the scheduler to handle. Doing so will incentivize low alert quality, which is counterproductive to users of this application.

bfrizb: Good point. I'll  I'll update the pull request just to catch the SMTPException rather than all Exceptions

eastebry: I don't think that is necessarily what we want though. This part of the code base has no idea what the alerters beneath it are doing, so it doesn't really make sense to catch a specific error one of them might throw.

eastebry:
I am sort of thinking that we want to fix this higher up. Here is what I propose:
* create an "owner" field for each rule. 
* Modify handle_exception to email the stacktrace to the owner of the rule. 
* pokemon exception handling here: https://github.com/Yelp/elastalert/blob/master/elastalert/elastalert.py#L596 (as high up as possible

This should stop elastalert rules from taking down other rules. In addition:
* modify elastalert-test-rule to validate rules against a schema

ivanlei:
I like @eastebry's suggestion but I would say this should also work if there is no owner, or the format of the owner is invalid. There should be some default place it emails in that case.

elastalert/elastalert.py
Qmando: I think it would be nice to have some kind of smart exception handling, so that an unexpected exception can cause ElastAlert to disable the specific rule that threw that exception and continue with the rest. Almost all bugs we've found trigger every time a specific alert/rule runs, and this would cause the alert/rule to continually throw exceptions every time ElastAlert runs.

PR #102, opened 2015-06-01
Adds JSON Schema for ruletypes files

talyian:
This ensures that the types of fields in rule definitions are correct. The schema is checked:
1. when elastalert starts and `load_config` is called
2. when `elastalert-test-rule` is run

`elastalert-test-rule` now also takes an option '--schema-only' to test the rule schema without hitting ElasticSearch for data.


elastalert/schema.yaml
```diff
@@ -0,0 +1,187 @@
...
+  alert_text_type: {enum: [alert_text_only, exclude_fields]}
+  timestamp_field: {type: string}
+  field: {}
+
+  ### Commands
+  command: *arrayOfString
+  pipe_match_json: {type: boolean}
+
+  ### Email
+  email: *arrayOfString
```
Qmando: This can also be just a string. I think single recipient is common enough that it should take both.

Qmando:
Looks good to me.

PR #109, opened 2015-06-03
Add options for connection timeout

bkochendorfer:
Just incase your query is really slow, you can bump the timeout in your config.yaml.


Qmando:
Thanks for adding this! Could you add a relevant section to the docs?

bkochendorfer:
@Qmando :+1: Good point, added.

PR #110, opened 2015-06-04
Added email notifications and disabling rules which throw exceptions

Qmando:
Wraps run_rule and alert (including enhancements) in catch-all exception handling. This will disable the rule by putting it into a separate disabled_rules list which are not run. When the rule is changed, it will be put back into the normal run queue.

This also adds the ability to set a global email for notifications about exceptions. It has the smtp_host, from_addr and reply_to options like the email alerter. These options can now be set globally which become the global defaults for the email alerter.

### DIFF
--- a/elastalert/config.py
+++ b/elastalert/config.py
@@ -59,6 +59,7 @@ def load_configuration(filename, conf=None):
     """ Load a yaml rule file and fill in the relevant fields with objects.
 
     :param filename: The name of a rule configuration file.
+    :param conf: The global configuration dictionary, used for populating defaults.
     :return: The rule configuration, a dictionary.
     """
     try:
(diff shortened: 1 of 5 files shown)


elastalert/config.py
```diff
@@ -55,7 +55,7 @@ def get_module(module_name):
     return module
 
 
-def load_configuration(filename):
+def load_configuration(filename, conf=None):
```
ivanlei: Please add a doc comment for `conf`

elastalert/config.py
```diff
@@ -67,12 +67,12 @@ def load_configuration(filename):
...
 
     rule['rule_file'] = filename
-    load_options(rule)
+    load_options(rule, conf)
     load_modules(rule)
     return rule
 
 
-def load_options(rule):
+def load_options(rule, conf=None):
```
ivanlei: Yeah, add that doc comment here too, please.

elastalert/elastalert.py
```diff
@@ -1046,6 +1071,42 @@ def handle_error(self, message, data=None):
...
+            tb = traceback.format_exc()
+            email_body += tb
+
+        if isinstance(self.notify_email, basestring):
+            self.notify_email = [self.notify_email]
+        email = MIMEText(email_body)
+        email['Subject'] = subject if subject else 'ElastAlert notification'
+        email['To'] = ', '.join(self.notify_email)
+        email['From'] = self.from_addr
+        email['Reply-To'] = self.conf.get('email_reply_to', email['To'])
```
talyian: should fallback to From

talyian:
lg2m

PR #111, opened 2015-06-05
Set aggregate_id to not_analyzed

### LINKED ISSUE #104: Aggregate alerts can include different alerts in them
By default, the field aggregate_id is analyzed. Because it's base64 encoded, it can be split into tokens when an id like "abcdefgh123-w" is generated. If another alert has an id of "w-987djshfsd", both will be sent together. This is because a query string is used instead of a term filter.

Qmando:
Fixes #104 by setting aggregate_id to not_analyzed


y-trobinso:
lgtm

PR #112, opened 2015-06-05
Prevent a new rule from being loaded if there is a name conflict

### LINKED ISSUE #101: Don't load new rules with conflicting names
ElastAlert checks for name conflicts when starting, but not when a new rule config is added while elastalert is running. This will cause problems when an alert fires, when one of the files changes, or when elastalert is restarted.

Qmando:
Prevent a new file from being adding at runtime with a conflicting name. Fixes #101

### DIFF
--- a/tests/base_test.py
+++ b/tests/base_test.py
@@ -497,6 +497,17 @@ def test_rule_changes(ea):
     mock_load.assert_any_call('rules/rule2.yaml')
     mock_load.assert_any_call('rules/rule3.yaml')
 
+    # A new rule with a conflicting name wont load
+    new_hashes = copy.copy(new_hashes)
+    new_hashes.update({'rule4.yaml': 'asdf'})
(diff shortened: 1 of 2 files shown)


tests/base_test.py
```diff
@@ -497,6 +497,17 @@ def test_rule_changes(ea):
     mock_load.assert_any_call('rules/rule2.yaml')
     mock_load.assert_any_call('rules/rule3.yaml')
 
+    # A new rule with a conflicting name wont load
+    new_hashes = copy.copy(new_hashes)
```
y-trobinso: Should this use copy.deepcopy?

y-trobinso:
lgtm

tests/base_test.py
Qmando: This dict is of the form {filename: hash, filename2: hash2, etc}. I think normal copy is ok in this case.

PR #113, opened 2015-06-05
Alert with missing query_key wont silence other values

### LINKED ISSUE #100: Alert with a missing query_key silences other alerts
If an alert occurs on a match with a missing query_key, a silence will be added for the entire rule, which overrides other alerts with different query_keys. A missing value for query_key should be replaced with some token representing a missing value.

Qmando:
Fixes #100 
Previously, if there was a match with a missing query_key value, it would silence every other alert. Now, it only silences other alerts that are also missing a query key.

Added a test which failed before the changes.

### DIFF
--- a/elastalert/elastalert.py
+++ b/elastalert/elastalert.py
@@ -460,7 +460,8 @@ class ElastAlerter():
                     key = '.' + str(match[rule['query_key']])
                 except KeyError:
                     # Some matches may not have a query key
-                    key = ''
+                    # Use a special token for these to not clobber all alerts
+                    key = '._missing'
             else:
                 key = ''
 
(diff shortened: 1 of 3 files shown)


elastalert/elastalert.py
```diff
@@ -460,7 +460,8 @@ def run_rule(self, rule, endtime, starttime=None):
                     key = '.' + str(match[rule['query_key']])
                 except KeyError:
                     # Some matches may not have a query key
-                    key = ''
+                    # Use a special token for these to not clobber all alerts
+                    key = '._missing'
```
y-trobinso: Make sure this is documented somewhere, in the possible event of namespace conflicts (ie, a real key named "._missing")

y-trobinso:
lgtm

PR #114, opened 2015-06-05
Lists and dicts are pretty-printed as json in alerts

### LINKED ISSUE #98: Allow JSON objects in alert to be expanded
If nested object are included in an alert, they are printed out as python dictionaries. ElastAlert should have the ability to automatically expand them into multiple lines instead of one line text blobs.

Qmando:
Fixes #98 

dicts and lists will be expanded into pretty-printed JSON in match strings.


elastalert/alerts.py
```diff
@@ -58,7 +59,19 @@ def basic_match_string(rule, match):
...
+                text += key + ': '
+                # Attempt to pretty-print primative objects as JSON
+                if type(value) in [list, dict]:
+                    try:
+                        text += simplejson.dumps(value, sort_keys=True, indent=4) + '\n'
+                    except TypeError:
+                        # Non serializable object, revert to str
+                        text += str(value) + '\n'
+                else:
+                    text += str(value) + '\n'
```
y-trobinso: You could probably do better here with string.format. The fact this appears in a function named "basic_match_string" is also somewhat dubious, and can be rewritten to better state the intent of this function.

jjsendor:
I have turned `basic_match_string` function into a class. In `JiraAlerter` its child class `JiraFormattedMatchString` is used to provide a JSON formatted output of all entries in `match` dictionary (apart from the entries starting with _top_events_) surrounded by `{code:json}` and `{code}` tags.

PR #120, opened 2015-06-08
top_count_number actually defaults to 5

### LINKED ISSUE #97: ElastAlert "top_count_number" no longer defaults to 5
bwbaugh noticed that "top_count_number" no longer defaults to 5. I'm not sure whether this is intentional and the docs were not updated, or if it's a minor bug in the code.

The old code (elastalert/ruletypes.py) use to be:
           number = self.rules.get('top_count_number', 5)
            extra = get_top_counts(map(lambda e: e[0], self.occurrences[key].data), top_count_keys, number)

Now it's (elastalert/elastalert.py):
counts = self.get_top_counts(rule, start, end, keys, rule.get('top_count_number'), qk)

As a result, the default value for get_top_counts is "None", resulting in all the top key values getting displayed.

Qmando:
Fixes #97 

None was being passed to the number=5 default parameter, clobbering it, and causing the default to come from terms_size, which defaults to 50. Now nothing is passed and get_top_counts has the logic to grab the default


y-trobinso:
lgtm

PR #121, opened 2015-06-09
Fixed buffer_time causing duplicate events when elastalert restarts.

### LINKED ISSUE #53: Duplicate events counted with buffer_time
When ElastAlert is restarted, it will ignore buffer_time on the first run and set starttime to the endtime of the last query run. However, on the 2nd query, it will again use buffer time, which can cause the query window to overlap the previous queries and data to be counted twice.

Qmando:
Also fixed query windows not lining up with use_count/terms_query.

Buffer_time fix: When ElastAlert starts and get_starttime returns a value, the first query would begin at that time but the subsequent ones would go back past that starttime. This could cause duplicate events to be counted with elastalert starts. This adds a minimum starttime if a previous run has been used.

Query windows: set_starttime estimated the query windows using now() - run_every, which was often off by milliseconds to several seconds. Now, the previous endtime is recorded and used for the new starttime. This causes subsequent queries to line up exactly.

Added tests for both of these cases.

Fixes #53

### DIFF
--- a/elastalert/elastalert.py
+++ b/elastalert/elastalert.py
@@ -389,15 +389,21 @@ class ElastAlerter():
             # Try to get the last run from elasticsearch
             last_run_end = self.get_starttime(rule)
             if last_run_end:
+                rule['minimum_starttime'] = last_run_end
                 rule['starttime'] = last_run_end
                 return
 
         # Use buffer for normal queries, or run_every increments otherwise
         buffer_time = rule.get('buffer_time', self.buffer_time)
         if not rule.get('use_count_query') and not rule.get('use_terms_query'):
-            rule['starttime'] = endtime - buffer_time
+            # If we started using a previous run, don't go past that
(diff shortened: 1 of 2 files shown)


elastalert/elastalert.py
```diff
@@ -481,6 +487,9 @@ def run_rule(self, rule, endtime, starttime=None):
             # Add it as an aggregated match
             self.add_aggregated_alert(match, rule)
 
+        # Mark this endtime for next run's start
+        rule['previous_endtime'] = endtime
```
y-trobinso: I'm not enthused by the choice of variable names here. `minimum_starttime`, `starttime`, `previous_endtime`, `endtime`, and `buffer_time` mix styles and don't state their intent very well at all. I feel that concern is orthogonal to the change you want to make here, however.

In the future, I recommend we clean up all of this intermixing of responsibility, first by better stating intent, and later by separating these logic into well-defined classes and interfacing.

y-trobinso:
Quibbles with variable naming and intermixing of responsibility aside, lgtm.

PR #123, opened 2015-06-09
Add cc and bcc to email

oweizman:


docs/source/ruletypes.rst
```diff
@@ -648,6 +648,10 @@ by the smtp server.
 ``from_addr``: This sets the From header in the email. By default, the from address is ElastAlert@ and the domain will be set
 by the smtp server.
 
+``cc``: This sets the CC header in the email. By default, this is left empty.
+
+``bcc``: This sets the BCC header in the email. By default, this is left empty.
```
Qmando: This is slightly misleading. It adds these as recipients but does not add them anywhere in the email headers

y-trobinso: Given minor ambiguities in the RFCs for email, I think it makes sense to have these comments state what they do with the MIMEText and SMTP objects, instead of relying on their implementation. In this way, your comments respect the Law of Demeter, which will ensure they continue to be correct should the implementation change out from under you.

In other words, call out what these do to MIMEText instead, and these comments should work fine.

Qmando:
LGTM

PR #124, opened 2015-06-09
Added smtp login option

### LINKED ISSUE #116: Email Alert add SMTP login options
We are using Mailgun as SMTP services, this cloud services need authentication.
Can you can add some configuration options like smtp_user/smtp_passwd?

Qmando:
Fixes #116 

Adds SMTP auth options. By setting ``smtp_auth_file`` to a file which contains user and password field, the SMTP connection will be authenticated.

### DIFF
--- a/docs/source/ruletypes.rst
+++ b/docs/source/ruletypes.rst
@@ -642,6 +642,9 @@ Optional:
 
 ``smtp_host``: The SMTP host to use, defaults to localhost.
 
+``smtp_auth_file``: The path to a file which contains SMTP authentication credentials. It should be YAML formatted and contain
+two fields, ``user`` and ``password``. If this is not present, no authentication will be attempted.
+
 ``email_reply_to``: This sets the Reply-To header in the email. By default, the from address is ElastAlert@ and the domain will be set
 by the smtp server.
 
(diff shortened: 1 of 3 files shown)


docs/source/ruletypes.rst
```diff
@@ -642,6 +642,9 @@ Optional:
 
 ``smtp_host``: The SMTP host to use, defaults to localhost.
 
+``smtp_auth_file``: The path to a file which contains SMTP authentication credentials. It should be YAML formatted and contain
+two fields, ``user`` and ``password``. If this is not present, no authentication will be attempted.
+
```
y-trobinso: How is a user handled without a password? I presume by "no authentication", this will attempt anonymous SMTP?

elastalert/alerts.py
```diff
@@ -205,8 +219,12 @@ def alert(self, matches):
 
         try:
             self.smtp = SMTP(self.smtp_host)
+            if 'smtp_auth_file' in self.rule:
+                self.smtp.login(self.user, self.password)
```
y-trobinso: In line with my question above, this works with anonymous SMTP, correct?

y-trobinso:
lgtm. I just have some questions about how this is handled, but I believe my assumptions are correct that this is handled by anonymous SMTP relaying when no credentials are specified.

docs/source/ruletypes.rst
Qmando: ``from_addr`` sets the source address and From header, and works with anonymous SMTP, which is what it's been doing until now.

PR #127, opened 2015-06-10
Fix smtp iussues

bogdangi:


setup.py
```diff
@@ -29,5 +29,6 @@
         'pyyaml',
         'jsonschema',
         'simplejson',
+        'functools32',
```
Qmando: Is this necessary?

bogdangi: For my clear Debian 7 (wheezy) - yes.

Qmando: I'm curious what is missing without this. From my understanding, only sockets with SSL support is required. I'm slightly hesitant to drop support for python2.6.

ivanlei: @bogdangi can you share what errors you get without `functools32`? I tested in a wheezy docker and don't seem to need this. My experiment was:
```shell
$ docker run -tt -i debian:wheezy-backports /bin/bash
```

And then in the container
```shell
$ apt-get update
$ apt-get install git-core python-pip
$ pip install tox
$ git clone http://github.com/Yelp/elastalert
$ cd elastalert/
$ tox -edevenv
$ make test
```
Tests pass. That might not be enough to really suss out requirements I suppose.

PR #129, opened 2015-06-10
Added url encoding to kibana4 links

### LINKED ISSUE #128: URL encode kibana4 links
They don't hyperlink correctly because of special characters like single quotes and commas

Qmando:
Fixes #128 

Adds a call to urllib.quote for the query parameters in kibana4 dashboard links.

### DIFF
--- a/elastalert/kibana.py
+++ b/elastalert/kibana.py
@@ -1,4 +1,6 @@
 # -*- coding: utf-8 -*-
+import urllib
+
 from util import EAException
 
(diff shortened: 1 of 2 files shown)


y-trobinso:
Would it make sense to add tests for this change?

ivanlei:
I agree with @y-trobinso. Tests are a good idea.
At least let us know you tried this with an ugly K4 link and it worked.

PR #136, opened 2015-06-17
Segmenting is different for count queries versus normal

### LINKED ISSUE #135: Slow performance with large buffer_time
Rules with buffer_time should never have queries segmented smaller than buffer_time.

Qmando:
Two changes:

- If `buffer_time` is less than the time elapsed since the previous query end time, use the previous `endtime`. Previously, if you set `buffer_time` less than `run_every`, each query would be skipping the difference. This would also happen if total query time was greater than `run_every`.

- Fixes #135. Segmenting is done by either `buffer_time` or `run_every`, depending on the type of query. This mimicks the query size of normal operation, where buffer_time would be used for normal queries and `run_every` would be used for `use_count`/`terms_query`

### DIFF
--- a/elastalert/elastalert.py
+++ b/elastalert/elastalert.py
@@ -400,14 +400,15 @@ class ElastAlerter():
         # Use buffer for normal queries, or run_every increments otherwise
         buffer_time = rule.get('buffer_time', self.buffer_time)
         if not rule.get('use_count_query') and not rule.get('use_terms_query'):
+            buffer_delta = endtime - buffer_time
             # If we started using a previous run, don't go past that
-            if 'minimum_starttime' in rule and rule['minimum_starttime'] > endtime - buffer_time:
+            if 'minimum_starttime' in rule and rule['minimum_starttime'] > buffer_delta:
                 rule['starttime'] = rule['minimum_starttime']
             # If buffer_time doesn't bring us past the previous endtime, use that instead
-            elif 'previous_endtime' in rule and rule['previous_endtime'] > endtime - buffer_time:
+            elif 'previous_endtime' in rule and rule['previous_endtime'] > buffer_delta:
                 rule['starttime'] = rule['previous_endtime']
             else:
-                rule['starttime'] = endtime - buffer_time
+                rule['starttime'] = buffer_delta
         else:
             # Query from the end of the last run, if it exists, otherwise a run_every sized window
             rule['starttime'] = rule.get('previous_endtime', endtime - self.run_every)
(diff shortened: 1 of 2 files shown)


elastalert/elastalert.py
```diff
@@ -403,6 +403,9 @@ def set_starttime(self, rule, endtime):
             # If we started using a previous run, don't go past that
             if 'minimum_starttime' in rule and rule['minimum_starttime'] > endtime - buffer_time:
```
jjsendor: it would make sense to extract `endtime - buffer_time` to a variable as it is used 3x in the subsequent lines. Call it sth with `_delta` at the end, I guess that would make sense :)

elastalert/elastalert.py
```diff
@@ -440,11 +443,16 @@ def run_rule(self, rule, endtime, starttime=None):
...
 
-        # Run the rule
-        # If querying over a large time period, split it up into chunks
+        # Run the rule. If querying over a large time period, split it up into segments
+        # The segment size is either buffer_size for normal queries or run_every for
+        # count style queries
         self.num_hits = 0
-        while endtime - rule['starttime'] > self.run_every:
-            tmp_endtime = rule['starttime'] + self.run_every
+        if not rule.get('use_count_query') and not rule.get('use_terms_query'):
```
jjsendor: How about extracting this if/else to a separate method? It could have 2 returns (thus no need for `else`) and `segment_size` could take a value of what is returned from that method.

tests/base_test.py
```diff
@@ -411,22 +411,35 @@ def test_queries_with_rule_buffertime(ea):
         mock_es_init.return_value = mock_es
         ea.run_rule(ea.rules[0], END, START)
 
-    # Assert that es.search is run against every run_every timeframe between START and END
+    # Assert that es.search is run against every buffer_time timeframe between START and END
```
jjsendor: My suggestion here would be to extract this to a method and provide a nice long name rather than a comment ;)

tests/base_test.py
```diff
@@ -411,22 +411,35 @@ def test_queries_with_rule_buffertime(ea):
...
         ea.current_es.search.assert_any_call(body=query, size=ea.max_query_size, index='idx', ignore_unavailable=True, _source_include=['@timestamp'])
 
     # Assert that num_hits correctly summed every result
     assert ea.num_hits == ea.current_es.search.call_count
 
 
+def test_query_segmenting(ea):
+    # Test that with use_count_query is set, segmenting occurs by run_every instead of buffer_time
+    ea.rules[0]['use_count_query'] = True
+    with mock.patch('elastalert.elastalert.Elasticsearch'):
```
jjsendor: Use `contextlib.nested` instead of multiple `with`s.

y-trobinso:
I continue to wish we had more distinct names for our various time-related variables (starttime, minimum_starttime, etc). I might try to tackle that on my own at some point.

Other than that and the existing comments, lgtm.

Issue #66, opened 2015-04-16
Don't comment on closed JIRA tickets


Qmando:
ElastAlert should open a new ticket instead of commenting on a ticket that has already been closed.

PR #137, opened 2015-06-17
Fix for Issue #66 - "Don't comment on closed JIRA tickets"

Javaru:
I added two configuration options, `jira_bump_not_in_statuses` and `jira_bump_in_statuses`, to allow for also checking the tickets status when the bump jira query is executed. I added appropriate documentation to the `ruletypes.rst` documentation. 

I did not set a default values to give the user the option checking either for tickets in a status, or tickets not in a status. Additionally, not setting a default will prevent breaking backwards compatibility issues.


Qmando:
Thanks for doing this! Looks good to me.

Issue #134, opened 2015-06-15
Add mechanism for enhancement to drop a match


Qmando:
This could come in the form of another exception type which will be tell elastalert to drop the match without logging it as an error.

PR #138, opened 2015-06-17
Bfrisbie/134/enhancement drop match

bfrizb:
This pull request addresses https://github.com/Yelp/elastalert/issues/134 by adding a mechanism for enhancements to drop a match.

All tests pass for "make test"

I manually tested the changes by adding a simple enhancement to an existing rule that would raise DropMatchException() if match['username'] = 'baduser'.

$ python -m elastalert.elastalert --debug --start 2015-06-16T20:40:55Z --rule test_rule.yaml

Without enhancement:
INFO:root:Skipping writing to ES: {'rule_name': 'Test rule.baduser', '@timestamp': '2015-06-17T20:18:04.233412Z', 'exponent': 0, 'until': '2015-06-17T21:18:04.233401Z'}
INFO:root:1 match(es)
INFO:root:baduser matched Test rule at 2015-06-16T20:43:29Z

With enhancement:
INFO:root:Skipping writing to ES: {'rule_name': 'Test rule.baduser', '@timestamp': '2015-06-17T20:18:04.233412Z', 'exponent': 0, 'until': '2015-06-17T21:18:04.233401Z'}
INFO:root:0 match(es)
INFO:root:Ignoring match for silenced rule Test rule.baduser


elastalert/elastalert.py
```diff
@@ -819,11 +820,16 @@ def send_alert(self, matches, rule, alert_time=None):
...
+            valid_matches = []
             for match in matches:
                 try:
                     enhancement.process(match)
+                    valid_matches.append(match)
+                except DropMatchException as e:
+                    pass
                 except EAException as e:
                     self.handle_error("Error running match enhancement: %s" % (e), {'rule': rule['name']})
+            matches = valid_matches
```
Qmando: I think this function should return here if there are no matches left. Calling alerter.alert([]) is *probably* most of the time, but unfortunately I think there are a few places where matches[0] is referenced explicitly. Also, it looks like it would still write back to Elasticsearch that an empty alert was sent.

bfrizb: Done

PR #139, opened 2015-06-18
Test with mock data

Qmando:
This change adds the following features to this test script:

- Counts the number of documents from the last X days by using a single query to grab doc_type and fields, and then a count query to get a number, (and then up to 10,000 documents to save)
- Save data as JSON, for offline testing purposes. It will grab up to 10,000 documents form the last X days.
- Use offline data and mock out Elasticsearch for running against
- Actually runs ElastAlert using either a debug alerter or the real alert against real or file provided data
- Intercepts documents that would be written back to Elasticsearch

Other notes:

- It only takes one file at a time now
- The mock functions work for basic cases but may differ slightly sometimes.
- Some rules, like spike, may not trigger because elastalert is only run over the time period of the documents.
- Run_every affects the resolution of count and terms query, and it's set to the default 5 minutes here.
- This slightly changes the log text for when alerts are triggered

I also added a few various small changes in the docs to fix/clarify things

### DIFF
--- a/elastalert/test_rule.py
+++ b/elastalert/test_rule.py
@@ -4,18 +4,31 @@ from __future__ import absolute_import
 from __future__ import print_function
 
 import datetime
+import random
+import re
+import string
 
 import argparse
+import mock
+import simplejson
 import yaml
 
+from elastalert.config import load_modules
 from elastalert.config import load_options
 from elastalert.elastalert import ElastAlerter
 from elastalert.elastalert import Elasticsearch
 from elastalert.util import lookup_es_key
+from elastalert.util import replace_hits_ts
 from elastalert.util import ts_now
+from elastalert.util import ts_to_dt
+
+
+data = []
 
 
 def print_terms(terms, parent):
+    """ Prints a list of flattened dictionary keys """
+    print("\nAvailable terms in first hit:")
     for term in terms:
         if type(terms[term]) != dict:
             print('\t', parent + term)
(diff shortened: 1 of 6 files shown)


elastalert/test_rule.py
```diff
@@ -4,18 +4,31 @@
...
 from elastalert.config import load_options
 from elastalert.elastalert import ElastAlerter
 from elastalert.elastalert import Elasticsearch
 from elastalert.util import lookup_es_key
+from elastalert.util import replace_hits_ts
 from elastalert.util import ts_now
+from elastalert.util import ts_to_dt
+
+
+data = []
```
y-trobinso: Instead of making this a global, let's promote this and all of the functions that follow to a class and make `data` a dynamic member of said class.

elastalert/test_rule.py
```diff
@@ -4,18 +4,31 @@
 from __future__ import print_function
```
y-trobinso: Note that py3 __future__ imports have weird interactions with non-future printing. In line with the below, should someone attempt to use test_rule as a library, they should be aware of it.

elastalert/test_rule.py
```diff
@@ -23,63 +36,244 @@ def print_terms(terms, parent):
...
+    """ Loads a rule config file, performs a query over the last day (args.days), lists available keys
+    and prints the number of results. """
+    filename = args.file
+    with open(filename) as fh:
+        conf = yaml.load(fh)
+    load_options(conf)
+    print("Successfully loaded %s\n" % (conf['name']))
+
+    if args.schema_only:
+        return []
```
y-trobinso: What's the intent of this exit scenario? This function is named "check_files", yet schema_only... does not check files?

elastalert/test_rule.py
```diff
@@ -23,63 +36,244 @@ def print_terms(terms, parent):
...
+    query = ElastAlerter.get_query(conf['filter'], starttime=start_time, endtime=end_time, timestamp_field=ts)
+    index = ElastAlerter.get_index(conf, start_time, end_time)
+
+    # Get one document for schema
+    try:
+        res = es_client.search(index, size=1, body=query, ignore_unavailable=True)
+    except Exception as e:
+        print("Error running your filter:")
+        print(repr(e)[:2048])
+        return None
```
y-trobinso: Why print instead of stderr?

elastalert/test_rule.py
```diff
@@ -23,63 +36,244 @@ def print_terms(terms, parent):
...
+
+    # Get a count of all docs
+    count_query = ElastAlerter.get_query(conf['filter'], starttime=start_time, endtime=end_time, timestamp_field=ts, sort=False)
+    count_query = {'query': {'filtered': count_query}}
+    try:
+        res = es_client.count(index, doc_type=doc_type, body=count_query, ignore_unavailable=True)
+    except Exception as e:
+        print("Error querying Elasticsearch:")
+        print(repr(e)[:2048])
+        return None
```
y-trobinso: Same as above, in re, stderr.

elastalert/test_rule.py
```diff
@@ -23,63 +36,244 @@ def print_terms(terms, parent):
...
+    print("Got %s hits from the last %s day%s" % (num_hits, args.days, 's' if args.days > 1 else ''))
+    print_terms(terms, '')
+
+    # Check for missing keys
+    pk = conf.get('primary_key')
+    ck = conf.get('compare_key')
+    if pk and not lookup_es_key(terms, pk):
+        print("Warning: primary key %s is either missing or null!")
+    if ck and not lookup_es_key(terms, ck):
+        print("Warning: compare key %s is either missing or null!")
```
y-trobinso: Perhaps logging.info is more appropriate here? (And in line with my concerns about printing errors on stderr, we could use logging.error in those cases for consistency.)

elastalert/test_rule.py
```diff
@@ -23,63 +36,244 @@ def print_terms(terms, parent):
...
+    include = conf.get('include')
+    if include:
+        for term in include:
+            if not lookup_es_key(terms, term) and '*' not in term:
+                print("Included term %s may be missing or null" % (term))
+
+    for term in conf.get('top_count_keys', []):
+        if term not in terms:
+            print("top_count_key %s may be missing" % (term))
+    print('')
```
y-trobinso: What is this print doing, exactly?

elastalert/test_rule.py
```diff
@@ -23,63 +36,244 @@ def print_terms(terms, parent):
...
+            docs.append(doc)
 
-    for filename in args.files:
-        with open(filename) as fh:
-            conf = yaml.load(fh)
-        load_options(conf)
-        print("Successfully loaded %s\n" % (conf['name']))
+    # Remove all fields which don't match 'include'
+    for doc in docs:
+        for field in doc.keys():
```
y-trobinso: `for field in doc`. The key lookup is implicit.

elastalert/test_rule.py
```diff
@@ -23,63 +36,244 @@ def print_terms(terms, parent):
...
+            if field != '_id':
+                if not any([re.match(incl.replace('*', '.*'), field) for incl in rule['include']]):
+                    doc.pop(field)
 
-        if args.schema_only:
+    # Separate _source and _id, convert timestamps
+    resp = [{'_source': doc, '_id': doc['_id']} for doc in docs]
+    [doc['_source'].pop('_id') for doc in resp]
```
y-trobinso: Split this over two lines. Mutators of other data structures in the body of a list comprehension is not good style.

elastalert/test_rule.py
```diff
@@ -23,63 +36,244 @@ def print_terms(terms, parent):
...
+    replace_hits_ts(resp, rule)
+    return resp
+
+
+def mock_terms(rule, start, end, index, key, qk=None, size=None):
+    """ Mocks the effects of get_hits_terms using global data instead of Elasticsearch. """
+    if key.endswith('.raw'):
+        key = key[:-4]
```
y-trobinso: `key.rsplit('.raw', 1)[0]` will also do the trick. But given the explicit use of `.endswith` here, chopping off the last four characters is a safe operation.

elastalert/test_rule.py
```diff
@@ -23,63 +36,244 @@ def print_terms(terms, parent):
...
+    load_options(rule)
+    load_modules(rule)
+    conf['rules'] = [rule]
+
+    # If using mock data, make sure it's sorted and find appropriate time range
+    timestamp_field = rule.get('timestamp_field', '@timestamp')
+    if args.json:
+        global data
+        if not data:
+            return
```
y-trobinso: If we turn this into a class, the need to global everywhere goes away. :)

elastalert/test_rule.py
```diff
@@ -23,63 +36,244 @@ def print_terms(terms, parent):
...
+        try:
+            data.sort(key=lambda x: x[timestamp_field])
+            starttime = ts_to_dt(data[0][timestamp_field])
+            endtime = data[-1][timestamp_field]
+            endtime = ts_to_dt(endtime) + datetime.timedelta(seconds=1)
+        except KeyError as e:
+            print("All documents must have a timestamp and _id", e)
+            return
```
y-trobinso: Should this be on stderr?

elastalert/test_rule.py
```diff
@@ -23,63 +36,244 @@ def print_terms(terms, parent):
...
+
+        # Create mock _id for documents if it's missing
+        def get_id():
+            return ''.join([random.choice(string.letters) for i in range(16)])
```
y-trobinso: Are collisions a problem here?

elastalert/test_rule.py
```diff
@@ -23,63 +36,244 @@ def print_terms(terms, parent):
...
+
+        [doc.update({'_id': doc.get('_id', get_id())}) for doc in data]
```
y-trobinso: As before, please split this out instead of putting a mutator in the body of a list comprehension.

elastalert/test_rule.py
Qmando: check_file makes more sense. It used to take a list of files, but now it only checks one at a time.

elastalert/test_rule.py
```diff
@@ -1,86 +1,285 @@
...
+        end_time = ts_now()
+        ts = conf.get('timestamp_field', '@timestamp')
+        query = ElastAlerter.get_query(conf['filter'], starttime=start_time, endtime=end_time, timestamp_field=ts)
+        index = ElastAlerter.get_index(conf, start_time, end_time)
+
+        # Get one document for schema
+        try:
+            res = es_client.search(index, size=1, body=query, ignore_unavailable=True)
+        except Exception as e:
+            logging.info("Error running your filter:")
```
y-trobinso: Not sure if .info is appropriate here (I'd prefer if we got stderr), but I can see why this is reasonable given how the tool is invoked.

elastalert/test_rule.py
```diff
@@ -1,86 +1,285 @@
...
+        include = conf.get('include')
+        if include:
+            for term in include:
+                if not lookup_es_key(terms, term) and '*' not in term:
+                    logging.info("Included term %s may be missing or null" % (term))
+
+        for term in conf.get('top_count_keys', []):
+            # If the index starts with 'logstash', fields with .raw will be available but won't in _source
+            if term not in terms and not (term.endswith('.raw') and term[:-4] in terms and index.startswith('logstash')):
+                logging.info("top_count_key %s may be missing" % (term))
```
y-trobinso: I wish we had a schema for this instead of treating it like a special snowflake. Perhaps in a future change?

elastalert/test_rule.py
```diff
@@ -1,86 +1,285 @@
...
+        logging.info('')
+
+        # Download up to 10,000 documents to save
+        if args.save and not args.count:
             try:
-                res = es_client.search(index, size=1000, body=query)
+                res = es_client.search(index, size=10000, body=query, ignore_unavailable=True)
```
y-trobinso: Oh good. Now it's over 9000: https://www.youtube.com/watch?v=8GFnOGPGoOA

y-trobinso:
Small comments aside, lgtm. Thank you for moving this logic into a class, and for defining data properly on \_\_init\_\_.

Ship it!

PR #140, opened 2015-06-18
Update config.yaml.example

bkochendorfer:
Add some information about alerts. I ran into an issue trying to figure out how to add more information to the alert. I'm not sure this is the right place to put this but something like this would've helped me out.


Qmando:
I recommend adding this to examples/*.yaml. This file may get visibility, but it doesn't actually do anything in config.yaml. I am also planning on either highlighting this better in the documentation or changing it to be default behavior.

PR #141, opened 2015-06-19
Fixed elastalert_status reporting only the last segment of a query

Qmando:
If a query got segmented, only the last segment would be reported by as an elastalert_status. This changes the writeback body to use original_starttime instead of the (tmp) starttime


y-trobinso:
lgtm

PR #142, opened 2015-06-22
Fixed buffer_Time being ignored in favor of previous_endtime

Qmando:
buffer_time was ignored UNLESS it didn't go past previous_endtime. The test didn't catch this because use_count_query had been set by a previous test.

Verified this test failed after I popped use_count_query.


y-trobinso:
lgtm, based on separate discussion

PR #143, opened 2015-06-23
Compound query key

Qmando:
query_key can be set to a list, which will cause a compound key to be generated in the document by joining all fields by commas. The actual value of query key will then be set to this compound field name, which will exist in every document. If a value is missing, it will be treated as "None".

For example, with
```
query_key:
 - "username"
 - "computer"
```
and document 
`` {"username": "bob", "computer": "server1"} ``

query_key will be set to "username,computer" and the document will now include ``{"username,computer": "bob, server1"}`` so that the rest of the code can use this compound key.

I put this code into a new function, along with the functionality of replace_hits_ts, which did similar things. 

I added tests and documentation. I changed a bit of other document around query_key that implied it could only be used for certain rule types. I also added an example rule which uses this compound key and new_term rule.

### DIFF
--- a/docs/source/ruletypes.rst
+++ b/docs/source/ruletypes.rst
@@ -209,11 +209,12 @@ that will be given the match dictionary and can modify it before it is passed to
 ``module.file.EnhancementName``. See :ref:`Enhancements` for more information. (Optional, list of strings, no default)
 
 ``query_key``: Having a query key means that realert time will be counted separately for each unique value of ``query_key``. For rule types which
-count documents, such as spike, frequency and flatline, it also means that these counts will independent for each unique value of ``query_key``.
+count documents, such as spike, frequency and flatline, it also means that these counts will be independent for each unique value of ``query_key``.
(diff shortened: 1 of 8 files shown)


docs/source/ruletypes.rst
```diff
@@ -208,6 +208,13 @@ be converted to UTC, which is what ElastAlert uses internally. (Optional, boolea
 that will be given the match dictionary and can modify it before it is passed to the alerter. The enhancements should be specified as
 ``module.file.EnhancementName``. See :ref:`Enhancements` for more information. (Optional, list of strings, no default)
 
+``query_key``: Having a query key means that realert time will be counted separately for each unique value of ``query_key``. For rule types which
+count documents, such as spike, frequency and flatline, it also means that these counts will independent for each unique value of ``query_key``.
```
y-trobinso: It seems this line accidentally a word: "will [be] independent"

y-trobinso:
This seems reasonable enough to me. Do note that query key will become very confused if the string being matched happens to contain a comma. Given the nature of the keys you ingest, that seems unlikely but worth documenting as part of the contract.

Otherwise: ship it!

PR #148, opened 2015-06-24
Changed EventWindow.count to use a running sum

Qmando:
EventWindow.count() previous summed the counts of all events each time it ran, which was very inefficient considering it is often called after every append. This changes EventWindow to use a running count which is updated on insertion and deletion.

This also fixes a test which has been incorrectly using EventWindow.

I ran a small performance test which appended 10,000 events with random timestamps and called count() after each append. These changes cut the runtime roughly in half.


y-trobinso:
lgtm

PR #150, opened 2015-06-24
Handling _source not enabled and different timestamp_field

avanishp:
Agenda: 
Handling _source not enabled and different timestamp_field

Support for following use-case:

- timestamp is not as @timestamp but as _timestamp.
- _source is not enabled .
- timestamp is stored as long in ES
- smtp was not working with the default port

Changes done to handle these: 
- added the timestamp_field in the config file.
- added the timestamp_field in the fields part of the query and checking that instead of _source.
- added the timestamp_type , _source_enabled for above mentioned checks.
- added a few util methods to get the int part of the timestamp from a given timestamp.
- added smtp_port for the port and used ehlo , starttls methods to establish connection.


elastalert/elastalert.py
```diff
@@ -310,16 +325,20 @@ def remove_duplicate_events(self, data, rule):
...
         # Anything older than the buffer time we can forget
         now = ts_now()
         remove = []
         buffer_time = rule.get('buffer_time', self.buffer_time)
+        if rule.get('timestamp_type','datetime') == 'long':
+            now = dt_to_int(now)
+            buffer_time = timedelta_to_int(buffer_time)
         for _id, timestamp in rule['processed_hits'].iteritems():
+            if rule.get('timestamp_type','datetime') == 'long':
+                timestamp = dt_to_int(timestamp)
```
Qmando: Why convert now, buffer_time and timestamp to ints, when ``now - timestamp > buffer_time`` will still have the same results?

elastalert/elastalert.py
```diff
@@ -414,7 +432,7 @@ def set_starttime(self, rule, endtime):
             if 'minimum_starttime' in rule and rule['minimum_starttime'] > buffer_delta:
                 rule['starttime'] = rule['minimum_starttime']
             # If buffer_time doesn't bring us past the previous endtime, use that instead
-            elif 'previous_endtime' in rule and rule['previous_endtime'] < buffer_delta:
+            elif 'previous_endtime' in rule and rule['previous_endtime'] > buffer_delta:
```
Qmando: Is this because you have not pulled master since https://github.com/Yelp/elastalert/pull/142?

elastalert/config.py
```diff
@@ -108,12 +108,15 @@ def load_options(rule, conf=None):
     rule.setdefault('realert', datetime.timedelta(seconds=0))
     rule.setdefault('aggregation', datetime.timedelta(seconds=0))
     rule.setdefault('query_delay', datetime.timedelta(seconds=0))
-    rule.setdefault('timestamp_field', '@timestamp')
+    rule.setdefault('timestamp_field' , rule.get('timestamp_field' , '@timestamp'))
```
Qmando: This change doesn't do anything. setdefault doesn't have any effect if timestamp_field is already in the dictionary.

elastalert/alerts.py
```diff
@@ -219,7 +217,12 @@ def alert(self, matches):
 
         try:
             self.smtp = SMTP(self.smtp_host)
+            if(self.smtp_port):
+                self.smtp = SMTP(self.smtp_host,self.smtp_port)
+                self.smtp.ehlo()
+                self.smtp.starttls()
```
Qmando: This will throw an exception if starttls is not supported by the server. Specifying the smtp_port should not mean always use starttls. ehlo() will tell you if STARTTLS is supported, but it is not checked here.

elastalert/elastalert.py
```diff
@@ -197,21 +205,24 @@ def get_index_start(self, index, timestamp_field='@timestamp'):
         """
         query = {'sort': {timestamp_field: {'order': 'asc'}}}
         try:
-            res = self.current_es.search(index=index, size=1, body=query, _source_include=[timestamp_field], ignore_unavailable=True)
+            res = self.current_es.search(index=index, size=1, body=query, _source_include=[timestamp_field] ,ignore_unavailable=True)
```
Qmando: Why was this changed? Please use func(arg, arg2) format instead of func(arg , arg2).

elastalert/elastalert.py
```diff
@@ -224,9 +235,13 @@ def get_hits(self, rule, starttime, endtime, index):
         :param endtime: The latest time to query.
         :return: A list of hits, bounded by self.max_query_size.
         """
-        query = self.get_query(rule['filter'], starttime, endtime, timestamp_field=rule['timestamp_field'])
+        query = self.get_query(rule['filter'], starttime, endtime, timestamp_field=rule['timestamp_field'] , timestamp_type = rule['timestamp_type'])
         try:
-            res = self.current_es.search(index=index, size=self.max_query_size, body=query, _source_include=rule['include'], ignore_unavailable=True)
+            if rule['_source_enabled']:
+                res = self.current_es.search(index=index, size=self.max_query_size, _source_include=rule['include'] , body=query, ignore_unavailable=True)
+                logging.info(res)
```
Qmando: This should probably be logging.debug

avanishp:
Thanks for your comments. 
Will surely check all these comments and modify the code.

elastalert/elastalert.py
avanishp: I have not changed it and just merged it with the master of Yelp/elastalert. Please let me know if i have to revert this.

avanishp: Reverted this change as per #142

avanishp:
Hi Team,

Any updates on this Pull Request?

Qmando:
avanishp: I've made some changes to your code and will publish it shortly.

Qmando:
@avanishp: Please check out #160. I've changed a few things from your PR such as naming the timestamp types "unix", "unix_ms" and "iso". In addition, fields are supported but internally it's merged into _source. Please check it out and let me know if it works for you.

avanishp:
I saw the code. Have even left a comment there as the _source_include part in the es query will never return any result if the _source is not enabled in the ES.

PR #151, opened 2015-06-24
Changed logging back to print, fixed replace_hits_ts import error

Qmando:
1. Fixes replace_hits_ts which moved to process_hits in another PR.
2. Change all logging.info back to print/print to stderr. This is mostly because having "INFO:root:" preceding every line was annoying and unnecessary. I think print is appropriate for a test script.
3. buffer_time is changed to 45 minutes, which is the same as the config.yaml.example

### DIFF
--- a/elastalert/test_rule.py
+++ b/elastalert/test_rule.py
@@ -1,12 +1,14 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 from __future__ import absolute_import
+from __future__ import print_function
 
 import datetime
 import logging
 import random
 import re
 import string
+import sys
 
 import argparse
 import mock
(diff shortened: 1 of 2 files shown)


elastalert/test_rule.py
```diff
@@ -79,44 +81,44 @@ def test_file(self, args):
...
                 if not lookup_es_key(terms, term) and '*' not in term:
-                    logging.info("Included term %s may be missing or null" % (term))
+                    print("Included term %s may be missing or null" % (term), file=sys.stderr)
 
         for term in conf.get('top_count_keys', []):
             # If the index starts with 'logstash', fields with .raw will be available but won't in _source
             if term not in terms and not (term.endswith('.raw') and term[:-4] in terms and index.startswith('logstash')):
-                logging.info("top_count_key %s may be missing" % (term))
-        logging.info('')
+                print("top_count_key %s may be missing" % (term), file=sys.stderr)
```
y-trobinso: These technically work, though I'm curious why you didn't instead opt for [logging.error](https://docs.python.org/2/library/logging.html#logging.error), then direct ERROR and CRITICAL to stderr.

PR #158, opened 2015-06-26
All SSL connections to SMTP host

jcockhren:
Setting `smtp_ssl` to `true` in a rule allows for the alerter class to initiate an SSL connection. Very useful when using services like Mandrill and SendGrid.


CCoffie:
:+1:

PR #164, opened 2015-06-30
'Available terms' message printed once for recursive call

0xdabbad00:
Simple fix to move the print statement out of the recursive call to avoid having it print multiple times


Qmando:
Looks good to me

PR #167, opened 2015-07-01
Pass args to RuleType.__init__ and update NewTerm to obey --start

### LINKED ISSUE #166: new_term rule doesn't use --start
It always uses the current time to make queries, and without strftime index, it doesn't use any time range at all. It should use --start to bound it's window for testing purposes.

Qmando:
Fixes #166 

args gets passed to RuleType.__init__ so that rule types can use arguments. This also always adds a time range bound to get_all_terms in NewTermRule.

These changes are pretty expansive considering the small scope of the issue this is to solve. Maybe forcing load_rules to take a parsed arg object is not ideal. I am open to suggestions on better ways to do this.

### DIFF
--- a/elastalert/config.py
+++ b/elastalert/config.py
@@ -55,7 +55,7 @@ def get_module(module_name):
     return module
 
 
-def load_configuration(filename, conf=None):
+def load_configuration(filename, conf=None, args=None):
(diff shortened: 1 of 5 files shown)


y-trobinso:
lgtm

PR #174, opened 2015-07-10
Fixing reloading of rules.

CCoffie:
I'm not sure if you guys encountered this but the rules folder was appended twice so I would get something like the following `rules/rules/rule_name.yaml` where I just wanted `rules/rule_name.yaml`.

This fixed the problem. Tell me if this works for you guys.


Qmando:
Nice find! This seems to have gone unnoticed for so long because with full file paths, os.path.join will deduplicate the rules folder path.

Qmando:
I've fixed this in another place and fixed a test as well. Closing this PR for #175

PR #175, opened 2015-07-10
Fix rule file loading

Qmando:
rule_file paths already included the elastalert_rules folder. When they were joined again, this resulted in the duplication of the elastalert_rules path. os.path.join would de-duplicate this when elastalert_rules was given as a full path but not as a relative path.

This also fixes the test to have a full path as mock rule_hashes data.


y-trobinso:
lgtm after a short chat

PR #182, opened 2015-07-17
Add Vagrantfile and Salt provisioning

thomdixon:
This is an entirely optional merge, but I found this setup convenient, so perhaps Yelp  (or someone else) might too.

This branch adds a Vagrantfile which creates a multi-machine environment provisioned by Salt that consists of the following:

* `elasticsearch`:
    - This is an Ubuntu 14.04 box with Elasticsearch 1.6.
* `elastalert`:
    - This is an Ubuntu 14.04 box with:
        - ElastAlert and dependencies
        - Python 2.6 and 2.7
        - Supervisor

Upon instantiation, the `elastalert` box will be provisioned to use `elasticsearch` as its Elasticsearch host. Moreover, ElastAlert will be automatically started under `supervisord` using the configuration in `salt/roots/salt/elastalert/config.yaml` with the rules (if any) specified in `salt/roots/salt/elastalert/rules`. Combined with the scripted loading of data for the `elasticsearch` box, this can be a convenient way to test rules in an automated fashion without a persistent Elasticsearch environment.

The automatic configuration of `/etc/hosts` relies on the Vagrant plugin `vagrant-hostmanager`, and so this plugin is required in order to use this Vagrantfile.

Neither VirtualBox nor VMWare support hard links being created inside shared folders, and so `setup.py` had to be modified in order to prevent the creation of hard links during the run of `python setup.py sdist` performed by `tox`.

This PR depends on #179 to automate the initial index creation, and so it has been included in the history of this branch.

All tests pass for both Python 2.6 and 2.7 when running: `vagrant ssh elastalert -- "cd /vagrant; sudo tox"`

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,0 +1,38 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+# Vagrantfile API/syntax version. Don't touch unless you know what you're doing!
+VAGRANTFILE_API_VERSION = "2"
+
+unless Vagrant.has_plugin?("vagrant-hostmanager")
+  raise 'vagrant-hostmanager is not installed! '\
+        'Please run: vagrant plugin install vagrant-hostmanager'
+end
+
+Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
+  config.vm.box = "rubicon/ubuntu1404-salt"
+  config.vm.synced_folder "salt/roots/", "/srv/"
+
+  config.hostmanager.enabled = true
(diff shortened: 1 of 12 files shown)


Qmando:
This is great, but I think that it should be separate repository. I have no experience with Salt and don't think it really makes sense to maintain this along with ElastAlerts source. I would happily put a link in the docs to wherever this lives. I'll happily merge the create_index changes.

thomdixon:
@Qmando I'll make it a separate repo :)

PR #184, opened 2015-07-20
read input once, to support piping rules from stdin

talyian:
this reads the input only once, in case it is non-rewindable.

### DIFF
--- a/elastalert/test_rule.py
+++ b/elastalert/test_rule.py
@@ -3,6 +3,7 @@
 from __future__ import absolute_import
 from __future__ import print_function
 
+import copy
 import datetime
 import logging
 import random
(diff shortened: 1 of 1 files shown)


elastalert/test_rule.py
```diff
@@ -271,18 +267,21 @@ def run_rule_test(self):
...
 
+        with open(args.file) as fh:
+            rule_yaml = yaml.load(fh)
+
         if args.json:
             with open(args.json, 'r') as data_file:
                 self.data = simplejson.loads(data_file.read())
         else:
-            hits = self.test_file(args)
+            hits = self.test_file(copy.deepcopy(rule_yaml), args)
```
talyian: Validation is not idempotent, since it mutates the rule object, changing, for example, `{"hours": 4}` to a `datetime.timedelta`. Therefore, we have to create a deepcopy if we want to run the rule through validation twice.

PR #185, opened 2015-07-20
Specify Kibana4 dashboard timedelta

CCoffie:
Added the option to specify the kibana 4 dashboard timeframe. I also updated the docs to describe the new option.

### DIFF
--- a/elastalert/elastalert.py
+++ b/elastalert/elastalert.py
@@ -688,8 +688,8 @@ class ElastAlerter():
     def generate_kibana4_db(self, rule, match):
         ''' Creates a link for a kibana4 dashboard which has time set to the match. '''
         db_name = rule.get('use_kibana4_dashboard')
-        start = ts_add(match[rule['timestamp_field']], -rule.get('kibana4_start_timedelta', datetime.timedelta(minutes=10)))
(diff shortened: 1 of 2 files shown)


elastalert/elastalert.py
```diff
@@ -688,8 +688,8 @@ def sleep_for(self, duration):
     def generate_kibana4_db(self, rule, match):
         ''' Creates a link for a kibana4 dashboard which has time set to the match. '''
         db_name = rule.get('use_kibana4_dashboard')
-        start = ts_add(match[rule['timestamp_field']], -rule.get('timeframe', datetime.timedelta(minutes=10)))
-        end = ts_add(match[rule['timestamp_field']], datetime.timedelta(minutes=10))
+        start = ts_add(match[rule['timestamp_field']], -rule.get('kibana4_start_timedelta', datetime.timedelta(minutes=10)))
```
Qmando: Could this be
``-rule.get('kibana4_start_timedelta', rule.get('timeframe', datetime.timedelta(minutes=10)))``

That way, before falling back to 10 minutes, it will use the timeframe instead.

CCoffie: Just added that in. Tell me what you think.

PR #191, opened 2015-07-22
Changed query ranges to be only partially inclusive to avoid duplicates

### LINKED ISSUE #190: Possible double counted documents if they fall on query boundary
The time range for count and terms query is include for both lower and upper bounds. This means that if a document has the exact timestamp where a query window falls, it will be counted twice. 

Timestamps are of second precision and queries are not always exactly equally spaced apart.

Qmando:
Fixes #190


talyian:
lg2m. I feel inclusive lower-bound, exclusive upper-bound is more usual but not totally important

PR #192, opened 2015-07-22
Fixed windows not being cleared after a match

Qmando:
Fixes a bug introduced in #148 where the eventwindow deques would be cleared but running_count would not. This caused alerts to fire erroneously after the first.


y-trobinso:
lgtm

PR #193, opened 2015-07-22
elastalert-test-rule should respect ES config

thomdixon:
Use ElastAlert's configuration builder to respect the ES options specified in the individual rule file, and use ElastAlert's factory to construct the client used to connect to ES.


Qmando:
Nice! Can you remove the Elasticsearch import now that it's unused? The tests are failing with
"elastalert/test_rule.py:22:1: F401 'Elasticsearch' imported but unused"

thomdixon:
@Qmando Oops! Fixed. :)

PR #196, opened 2015-07-23
Allow alert text, subject and spike query key to be nested fields

Qmando:
Use lookup_es_key instead of document.get

### DIFF
--- a/elastalert/alerts.py
+++ b/elastalert/alerts.py
@@ -15,6 +15,7 @@ from jira.client import JIRA
 from jira.exceptions import JIRAError
 from staticconf.loader import yaml_loader
 from util import EAException
+from util import lookup_es_key
 from util import pretty_ts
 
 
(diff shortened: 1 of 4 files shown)


y-trobinso:
lgtm. Add tests maybe?

elastalert/alerts.py
```diff
@@ -34,7 +35,8 @@ def _add_custom_alert_text(self):
         alert_text = self.rule.get('alert_text', '')
         if 'alert_text_args' in self.rule:
             alert_text_args = self.rule.get('alert_text_args')
```
talyian: not from this CR, but .get('alert_text_args', []) or otherwise the next lines will raise

Qmando: This is in an if block to make sure it exists. Though someone could do alert_text_args: 5 or something dumb if it's not in the JSON schema

talyian:
lg2m

PR #197, opened 2015-07-24
Use args.day for Mock elastalerter

### LINKED ISSUE #157: elastalert-test-rule always uses 1 day
It will count and save data from args.days, but MockElastAlerter always uses 1 day.

Qmando:
Fixes #157


y-trobinso:
lgtm. Add tests maybe?

PR #201, opened 2015-08-02
Support for must_not filter as in ES

avanishp:


avanishp:
Any updates on this?

Qmando:
I'm hesitant to include another config option when you can just use
```
filter:
  - not:
      foo: bar
```

avanishp:
Thanks .. I did not notice that part in the document

PR #208, opened 2015-08-18
Fixed bug where changed rule woudlnt load if it failed to load previo…

Qmando:
Fixes a bug where if a rule gets added, and its initialization throws an exception, it is in self.rule_hashes but not self.rules. So if it changes, it will be "reloaded" but then it doesn't exist in self.rules so it doesn't get added (It normally replaces the old version with the new).

This also adds a test for this case and fixes the mock rules which were incomplete but weren't causing errors until I hit the bugged code.


talyian:
just dooo it

PR #212, opened 2015-08-20
Set current_es for pending alerts

Qmando:
send_pending_alerts is called before run_rule. What this means is that if ElastAlert starts and an aggregation alert is ready to be sent, and it contains top_count_keys, an exception will occur when it tries to search via self.current_es. This sets current_es as it does in run_rule.


y-trobinso:
lgtm, but please add tests

talyian:
lg2m

PR #213, opened 2015-08-20
use lookup_es_key() for compound_query_key

wataru-chocola:


Qmando:
This is likely not the only place this needs to be replaced in elastalert.py, but it's a step forward.

PR #217, opened 2015-08-20
Added max_aggregation

### LINKED ISSUE #180: Only 10 alerts are aggregated in emails
I have more than 10 alerts pending to be sent. Alerts are aggregated.
When the aggregation interval elapses, only 10 alerts are put in the email.

I specified a size=100 in get_aggregated_matches, that seems to solve the issue. I don't see a way to make this configurable.

Qmando:
Fixes #180 

Adds the max_aggregation setting, limiting the query in get_aggregated_matches


y-trobinso:
lgtm, but please add tests

PR #226, opened 2015-08-25
Removed default es_debug_trace

Qmando:
Removes the defualt value for ``es_debug_trace`` so that it can be disabled.


y-trobinso:
Should this change be in release notes, because it may impact defaults other people rely upon?

Otherwise, lgtm.

Qmando:
It's only been default since 0.0.53, so I doubt it will affect people much, especially since it's for debugging only.

y-trobinso:
Works for me.

PR #227, opened 2015-08-25
Added newline to cmd json pipe

### LINKED ISSUE #225: pipe_match_json does not seem to send json match to script
Using the following in a rules:
```
pipe_match_json: true
command: ["/etc/elastalert/script.sh"]
```

Here is my simple script:
```
#!/bin/bash

echo "New alert at $(date)" >> /var/log/elastalert.log

while read -r line; do
        echo "${line}" >> /var/log/elastalert.log
done < /dev/stdin
```

does not seem to pass anything to the stdin of the script. I can see an entry from the first 'echo' command but no JSON.

I try to run the script with the following command
```
echo -e 'blah\nand\nblah' | ./script.sh
```

And I confirm it logs stdin.

Qmando:
Fixes #225 

Without the newline, some commands would fail to read the json. An example of this is in the issue.

### DIFF
--- a/elastalert/alerts.py
+++ b/elastalert/alerts.py
@@ -434,7 +434,7 @@ class CommandAlerter(Alerter):
             subp = subprocess.Popen(command, stdin=subprocess.PIPE)
 
             if self.rule.get('pipe_match_json'):
-                match_json = json.dumps(matches)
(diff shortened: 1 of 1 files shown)


y-trobinso:
lgtm

PR #230, opened 2015-08-25
Added minimum_starttime to new rule copied fields

### LINKED ISSUE #229: Duplicate alert with after restarting then changing a rule
Because minimum_starttime is not copied with rules, if minimum_starttime is preventing a duplicate alert, altering the rule will remove it and cause another alert to fire.

Qmando:
Fixes #229 

Also adds a check to prevent keyerrors if the copy fields aren't present. The 3 others beside minimum_starttime and starttime should always be present anyway.


y-trobinso:
lgtm

PR #234, opened 2015-08-31
Now we don't log everything to root, we log it to an 'elastalert' logger

MattyKuzyk:


Qmando:
Hm, I don't really like having a mix of global variable and class properties. Maybe it would be better to have it be a global, perhaps in ``util.py``, and have ``from util import elastalert_logger`` in every other file.

elastalert/elastalert.py
```diff
@@ -60,6 +60,7 @@ def parse_args(self, args):
         self.args = parser.parse_args(args)
 
     def __init__(self, args):
+        self.elastalert_logger = logging.getLogger('elastalert')
```
Qmando: Can this file import the global logger as well?

PR #236, opened 2015-09-01
notify_email field will now have email sent to that address for slow …

MattyKuzyk:
…rules and rule exceptions

### DIFF
--- a/elastalert/elastalert.py
+++ b/elastalert/elastalert.py
@@ -640,6 +640,7 @@ class ElastAlerter():
                         raise EAException("A rule with the name %s already exists" % (new_rule['name']))
                 except EAException as e:
                     self.handle_error('Could not load rule %s: %s' % (rule_file, e))
+                    self.send_notification_email(exception=e, rule=new_rule)
                     continue
                 logging.info('Loaded new rule %s' % (rule_file))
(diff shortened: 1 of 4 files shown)


Qmando:
Can you add a call to  ``self.send_notification_email`` at line 642? (Exception when loading a new rule)

elastalert/elastalert.py
```diff
@@ -1182,6 +1194,8 @@ def send_notification_email(self, text='', exception=None, rule=None, subject=No
         email = MIMEText(email_body)
         email['Subject'] = subject if subject else 'ElastAlert notification'
         email['To'] = ', '.join(self.notify_email)
+        if rule and rule['notify_email']:
```
Qmando: Can you check here that ``rule['notify_email']`` is not the same as ``self.notify_email``?

Qmando:
I think the actual recipient addresses are on line 1204, ``smtp.sendmail(self.from_addr, self.notify_email, email.as_string())``, the rule's ``notify_email`` needs to be added there as well.

MattyKuzyk:
Yeah, it would seem as though the fields on `email` I edited are just headers that don't affect the actual sending. Weird.

PR #239, opened 2015-09-03
OpsGenie Alerter 

Ropes:
This PR adds a simple [OpsGenie](https://www.opsgenie.com/) Alerter in `modules/opsgenie.py`. It is a simple alerter which if matched by a rule sends an https request to [OpsGenie's API to create](https://www.opsgenie.com/docs/web-api/alert-api#createAlertRequest) an alert which can be directed at recipients for notification. Our use case is roughly; a log message that we don't like appears in Elasticsearch, notify team X that there might be an issue to investigate.

I think I updated the documentation in the relevant places but I might have missed something.

I've primarily been testing with a frequency rule, however once we get it deployed to staging more might find issues but the OGAlerter is pretty stupid simple. I did create a couple unit tests for the OpsGenieAlerter, however my python chops aren't what they used to be so critique welcome.

`tox` passes all tests on my end :four_leaf_clover:

### DIFF
--- a/docs/source/ruletypes.rst
+++ b/docs/source/ruletypes.rst
@@ -1015,8 +1015,6 @@ The OpsGenie alert requires four options:
 
 ``opsgenie_recipients``: An optional filled list OpsGenie recipients who will be notified by the alerts
 
-``opsgenie_addr``: Potentially unnecessary variable to configure the URL to POST alert requests to. It hasn't changed yet, but if it does not having it hardcoded seems advantageous.
-
 Debug
 ~~~~~~
 
--- a/docs/source/ruletypes.rst
+++ b/docs/source/ruletypes.rst
@@ -789,17 +789,19 @@ Example usage::
 OpsGenie
 ~~~~~~~~
 
-OpsGenie alerter will create an alert which can be used to notify Operations people of issues or log information. An OpsGenie ``API`` integration must be created in order to acquire the necessary ``opsgenie_key`` rule variable. Currently the OpsGenieAlerter only creates an alert, however it could be extended to update or close existing alerts.
+OpsGenie alerter will create an alert which can be used to notify Operations people of issues or log information. An OpsGenie ``API``
+integration must be created in order to acquire the necessary ``opsgenie_key`` rule variable. Currently the OpsGenieAlerter only creates 
+an alert, however it could be extended to update or close existing alerts.
 
-The alert requires four options:
+The OpsGenie alert requires four options:
 
-``opsgenie_key``: The randomly generated API Integration key create by OpsGenie.
+``opsgenie_key``: The randomly generated API Integration key created by OpsGenie.
 
 ``opsgenie_account``: The OpsGenie account to integrate with.
 
 ``opsgenie_recipients``: An optional filled list OpsGenie recipients who will be notified by the alerts
 
-``opsgenie_addr``: Potentially unnecessary variable to configure the URL to POST alert requests to. It hasn't changed yet, but if it does not having it hardcoded seems adventageous.
+``opsgenie_addr``: Potentially unnecessary variable to configure the URL to POST alert requests to. It hasn't changed yet, but if it does not having it hardcoded seems advantageous.
 
 Debug
 ~~~~~~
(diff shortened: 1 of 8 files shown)


example_rules/example_opsgenie_frequency.yaml
```diff
@@ -0,0 +1,67 @@
...
+- query:
+    query_string:
+      query: "@message: *hihi*"
+
+# (Required)
+# The alert is use when a match is found
+alert:
+- "opsgenie"
+
+# (required, email specific)
```
Qmando: This is only relevant if an email alerter is set. Either remove it or add ``- "email"`` to the alert section.

docs/source/ruletypes.rst
```diff
@@ -1000,6 +1000,23 @@ Example usage::
...
+
+The OpsGenie alert requires four options:
+
+``opsgenie_key``: The randomly generated API Integration key created by OpsGenie.
+
+``opsgenie_account``: The OpsGenie account to integrate with.
+
+``opsgenie_recipients``: An optional filled list OpsGenie recipients who will be notified by the alerts
+
+``opsgenie_addr``: Potentially unnecessary variable to configure the URL to POST alert requests to. It hasn't changed yet, but if it does not having it hardcoded seems advantageous.
```
Qmando: This seems a little confusing. Can you put what the default value is?

modules/opsgenie.py
```diff
@@ -0,0 +1,64 @@
...
+    '''Sends a http request to the OpsGenie API to signal for an alert'''
+    required_options = frozenset(['opsgenie_key', 'opsgenie_account', 'opsgenie_recipients'])
+
+    def __init__(self, *args):
+        super(OpsGenieAlerter, self).__init__(*args)
+
+        self.account = self.rule.get('opsgenie_account', 'genie')
+        self.api_key = self.rule.get('opsgenie_key', 'key')
+        self.recipients = self.rule.get('opsgenie_recipients', ['genies'])
+        self.to_addr = self.rule.get('opsgenie_addr', 'example.opsgenie.net')
```
Qmando: Is this a valid default address?

modules/opsgenie.py
```diff
@@ -0,0 +1,64 @@
...
+        post['source'] = 'ElastAlert'
+        post['tags'] = ['ElastAlert', self.rule['name']]
+        logging.debug(json.dumps(post))
+
+        try:
+            r = requests.post(self.to_addr, json=post)
+
+            logging.debug('request response: {0}'.format(r))
+            if r.status_code != 200:
+                logging.error("Error sending alert request to OpsGenie! {0}".format(r.json()))
```
Qmando: I think raise_for_status will get caught and logged again, so this logging line will cause a duplicate log message.

modules/opsgenie.py
```diff
@@ -0,0 +1,64 @@
...
+                r.raise_for_status()
+            logging.info("Alert sent to OpsGenie")
+        except Exception as err:
+            logging.error("Error sending alert to OpsGenie: {0}".format(err))
+            return
```
Qmando: Can this re-raise the exception as util.EAException? That will cause it to be caught in ElastAlert and 
1. Upload an elastalert_error document to Elasticsearch
2. Save the alert as unsent, where it can be retried
3. Logged eventually here https://github.com/Yelp/elastalert/blob/master/elastalert/elastalert.py#L1150

So I guess if you reraise as EAException you can also remove the logging.error from here.

Ropes:
@Qmando thank you for the feedback. I don't disagree on any of the issues you brought up, so I'll push a fix.

Ropes:
@Qmando I think I've found a good exception raising path for the OpsGenie Alert. This is what the output looks like when a bogus google.com path is given as the (now optional)`opsgenie_addr` field: 
```
python -m elastalert.elastalert --verbose --rule lytics_rules/freq_rule.yaml 
ERROR:root:Error querying for last run: TransportError(404, u'IndexMissingException[[elastalert_status] missing]')
INFO:urllib3.connectionpool:Starting new HTTP connection (1): 192.168.115.42
INFO:elastalert:Queried rule elastalerttestrule from 9-8 14:32 PDT to 9-8 15:17 PDT: 900 hits
INFO:requests.packages.urllib3.connectionpool:Starting new HTTPS connection (1): google.com
INFO:elastalert:Error response from https://google.com/hihi
 API Response: <Response [404]>
ERROR:root:Error while running alert opsgenie: Error sending alert: 404 Client Error: Not Found
```
This should appear if there's an issue of some connection issue with the request being sent. I think it strikes a decent balance of being informative without too much over-logging.

Let me know what you think.

Qmando:
Looks good. I changed the style on a few lines and merged it in here: #244

PR #241, opened 2015-09-06
hipchat alerter

doodyparizada:
send matches to a room in hipchat

@rounds/server


doodyparizada:
Not ready yet. opened by mistake

PR #245, opened 2015-09-11
Moved opsgenie into elastalert/ folder

Qmando:
It was creating separate packages for the modules and elastalert folders. Lets put them together. (for now.)


y-trobinso:
lgtm

PR #257, opened 2015-09-28
Fixed keyerror in aggregated alerts

Qmando:
There was a bug in which modifying a rule which had a pending aggregation would cause it to KeyError.

When a new aggregation starts, ``current_aggregate_id`` and ``aggregate_alert_time`` are both set. These are always set together, and so if one is true, the other is assumed to exist. However, only ``current_aggregate_id`` is copied while a rule is reloaded, thus, causing a KeyError.

This prevents the KeyError by copying ``aggregate_alert_time`` and also double-checking that it exists in the conditional, which shouldn't be needed anymore anyway.

### DIFF
--- a/elastalert/elastalert.py
+++ b/elastalert/elastalert.py
@@ -578,6 +578,7 @@ class ElastAlerter():
 
         copy_properties = ['agg_matches',
                            'current_aggregate_id',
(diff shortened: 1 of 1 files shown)


y-trobinso:
lgtm. You may want to update tests to cover this.

PR #259, opened 2015-09-29
Use lookup_es_key when creating silence stashes

Qmando:


y-trobinso:
lgtm. Should tests be updated to cover this?

PR #266, opened 2015-10-07
Added Slack and HipChat alerters.

bolshoy:
Added alerters for Slack and HipChat, with some handy options. Amended config, schema and docs.


Qmando:
Awesome stuff! Thanks so much for adding these. It all looks good to me.

PR #268, opened 2015-10-08
Use sortedlist for event window

Qmando:
EventWindow.append_middle was VERY slow. This should make it better. Haven't benchmarked yet.


talyian:
lg2m

PR #272, opened 2015-10-13
Fix message for Cardinality rule type

jaguasch:
If Cardinality rule is used with min_cardinality, an exception is raised because the alert message only refers to max_cardinality.

With this check at get_match_str function from CardinalityRule class, message is customised depending on the term used


pandujar:
:+1:

Issue #78, opened 2015-04-23
Inline alert configurations 


Qmando:
Instead of alerts reading their configuration from the top level of the configuration file, they should have inline configuration which would allow multiple alerts of the same type. For example:

```
alert:
- email:
    recipient: foo@bar.baz
- email:
    recipient: baz@foo.bar
```

bechtoldt:
+1

PR #276, opened 2015-10-14
Inline alerts

MattyKuzyk:
Alerts can now be configured inline. See #78

### DIFF
--- a/elastalert/config.py
+++ b/elastalert/config.py
@@ -243,7 +243,12 @@ def load_modules(rule, args=None):
 
         elif isinstance(alert, dict):
             alert_name = alert.keys()[0]
-            inline_alerts.append((alert[alert_name], alerts_mapping[alert_name] if alert_name in alerts_mapping else get_module(alert_name)))
+
+            # Each Inline Alert is a tuple, in the form (alert_configuration, alert_class_object)
+            if alert_name in alerts_mapping:
+                inline_alerts.append((alert[alert_name], alerts_mapping[alert_name]))
+            else:
+                inline_alerts.append((alert[alert_name], get_module(alert_name)))
 
             if not issubclass(inline_alerts[-1][1], alerts.Alerter):
                 raise EAException('Alert module %s is not a subclass of Alerter' % (alert))
--- a/elastalert/config.py
+++ b/elastalert/config.py
@@ -234,12 +234,12 @@ def load_modules(rule, args=None):
     if type(rule['alert']) != list:
         rule['alert'] = [rule['alert']]
     for alert in rule['alert']:
-        if type(alert) == str and alert in alerts_mapping:
+        if isinstance(alert, basestring) and alert in alerts_mapping:
             rule_alerts.append(alerts_mapping[alert])
-        elif type(alert) == dict and alert.keys()[0] in alerts_mapping:
+        elif isinstance(alert, dict) and alert.keys()[0] in alerts_mapping:
             rule_alerts.append(alerts_mapping[alert.keys()[0]])
         else:
-            rule_alerts.append(get_module(alert))
+            rule_alerts.append(get_module(alert.keys()[0] if isinstance(alert, dict) else alert))
             if not issubclass(rule_alerts[-1], alerts.Alerter):
                 raise EAException('Alert module %s is not a subclass of Alerter' % (alert))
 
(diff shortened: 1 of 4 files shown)


elastalert/config.py
```diff
@@ -233,13 +234,14 @@ def load_modules(rule, args=None):
     if type(rule['alert']) != list:
         rule['alert'] = [rule['alert']]
     for alert in rule['alert']:
-        if alert in alerts_mapping:
+        if type(alert) == str and alert in alerts_mapping:
```
Qmando: This should be ``if isinstance(alert, basestring) and ...`` so that it could work with unicode.

elastalert/config.py
```diff
@@ -251,7 +253,7 @@ def load_modules(rule, args=None):
 
     # Make sure we have required alert and type options
     reqs = rule['type'].required_options
-    for alert in rule['alert']:
+    for alert in rule_alerts:
         reqs = reqs.union(alert.required_options)
     if reqs - frozenset(rule.keys()):
```
Qmando: This looks like it will still throw an error if a required option is inline but not at the top level. I think the required_options checks have to be done per-rule now, probably after the copying happens.

elastalert/config.py
```diff
@@ -262,7 +264,17 @@ def load_modules(rule, args=None):
...
+        copied_alerts = copy.copy(rule['alert'])
+        rule['alert'] = []
+        for alert in copied_alerts:
+            if type(alert) == dict:  # For inline alert configs
+                for i, alert_conf in enumerate(alert.items()):
+                    copied_conf = copy.copy(rule)
+                    copied_conf.update(alert_conf[1])
+                    rule['alert'].append(rule_alerts[i](copied_conf))
+            else:
+                rule['alert'] = [alert(rule) for alert in rule_alerts]
```
Qmando: This should be
```
rule['alert'].append(alert(rule))
```

elastalert/config.py
```diff
@@ -262,7 +264,17 @@ def load_modules(rule, args=None):
         raise EAException('Error initializing rule %s: %s' % (rule['name'], e))
     # Instantiate alert
     try:
-        rule['alert'] = [alert(rule) for alert in rule['alert']]
+        copied_alerts = copy.copy(rule['alert'])
+        rule['alert'] = []
+        for alert in copied_alerts:
+            if type(alert) == dict:  # For inline alert configs
+                for i, alert_conf in enumerate(alert.items()):
```
Qmando: I'm not sure if we should use enumerate here. ``len(alert)`` here should generally be just 1 and ``rule_alerts`` may contain the old style alerts, so ``rule_alerts[i]`` will not always be right. I think ``rule_alert`` should be split into two lists, for inline and old style.

``inline_alerts`` could take the form of tuples of (AlertClass, alert_config) so that we don't need to enumerate and index, we can just go through each alert and instantiate it.

elastalert/config.py
```diff
@@ -233,13 +234,14 @@ def load_modules(rule, args=None):
     if type(rule['alert']) != list:
         rule['alert'] = [rule['alert']]
     for alert in rule['alert']:
-        if alert in alerts_mapping:
+        if type(alert) == str and alert in alerts_mapping:
             rule_alerts.append(alerts_mapping[alert])
+        elif type(alert) == dict and alert.keys()[0] in alerts_mapping:
+            rule_alerts.append(alerts_mapping[alert.keys()[0]])
```
Qmando: Along with the comment below, I think this should use a separate list for inline alerts. You can create a dictionary mapping because there could be multiple of the same type, IE, {Email: {...}, Email: {...}} wouldn't work because the alert type keys would conflict. Instead, we can create tuples.
```
rule_type = alerts_mapping[alert.keys()[0]]
rule_config = alert.values()[0]
inline_alerts.append((rule_type, rule_config))
```
We should probably also assert that len(alert) == 1, because we are doing direct indexing.

elastalert/config.py
```diff
@@ -229,17 +230,23 @@ def load_modules(rule, args=None):
...
-            if not issubclass(rule_alerts[-1], alerts.Alerter):
+        if isinstance(alert, basestring):
+            global_alerts.append(alerts_mapping[alert] if alert in alerts_mapping else get_module(alert))
+
+            if not issubclass(global_alerts[-1], alerts.Alerter):
+                raise EAException('Alert module %s is not a subclass of Alerter' % (alert))
+
+        elif isinstance(alert, dict):
+            alert_name = alert.keys()[0]
+            inline_alerts.append((alert[alert_name], alerts_mapping[alert_name] if alert_name in alerts_mapping else get_module(alert_name)))
```
Qmando: Can you split this into two lines?

PR #277, opened 2015-10-15
DOC: fix alert example code

Minjung:
basic_match_alert doesn't seem to exist anymore, instead there's a class BasicMatchString with its own __str__().


Qmando:
Thanks! I always forget about the examples when I make changes :/

PR #283, opened 2015-10-22
Added sleep to create_index

Qmando:
There was a race condition where the index would not be ready and a put_mapping call would error out. 2 seconds worked in testing but is arbitrary


bfrizb:
Probably would be good to add a comment indicating that time.sleep(2) is to avoid race conditions. Other, it looks okay to me.

PR #287, opened 2015-10-26
Update alert types to be consistent with other docs

jasonrhaas:


Qmando:
Thanks for doing this. I'm terrible at updating all the docs :(

PR #290, opened 2015-10-27
Add pagerduty

jaguasch:
Added pagerduty as an alerter.

It needs the following config options to be set at alert configuration file:
- pagerduty_client_name: Integration Key generated after creating a service with the 'Use our API directly' option at Integration Settings
- pagerduty_client_name: The name of the monitoring client that is triggering this event.

### DIFF
--- a/docs/source/ruletypes.rst
+++ b/docs/source/ruletypes.rst
@@ -81,7 +81,7 @@ Rule Configuration Cheat Sheet
 | ``_source_enabled`` (boolean, default True)                  |           |
 +--------------------------------------------------------------+-----------+
 
-| 
+|
 
 +----------------------------------------------------+-----+-----------+-----------+--------+-----------+-------+----------+--------+-----------+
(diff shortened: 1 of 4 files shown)


jasonrhaas:
nice!  After looking at the code quickly, do you think it might be better to make some of the PagerDuty API fields configurable?  For example, you have `description` pull the ElastAlert rule name, but I may want that to be something descriptive for people on call such as "Tweet Flatline:  No Tweets collected in the last 5 minutes."

jaguasch:
Thanks :)

For the moment, the whole alert message (the one received as the message in email type alerts) is on the 'details' object, in the 'information' key. I take the description field as the name because PagerDuty will use it for the incident title too.

I was trying to think about an improved version of this integration just adding as optional the rest of PagerDuty API fields (based on https://developer.pagerduty.com/documentation/integration/events/trigger), like client_url to reference the internal source service url, client_name (now included as mandatory) and the incident_key

pandujar:
:+1:

Qmando:
Looks good! Thanks for contributing this.

PR #294, opened 2015-10-29
Added jira_server to pipeline

Qmando:
Fixes a bug where ``jira_server`` was only added to pipeline when existing tickets were bumped


MattyKuzyk:
lgtm :+1:

PR #297, opened 2015-11-04
Added fallback for raw bytes in alert

Qmando:
If an object in the match contains non-unicode, it will crash when it attempts to json.dumps it. This will prevent that by falling back to decoding the string as Latin-1, which will always succeed.


MattyKuzyk:
:+1:

PR #304, opened 2015-11-09
Add capability to specify multiple indexes in a rule.

amazonDR:
Within our company we are using multiple clusters in elastic search. As a result our rules needed to support defining multiple indexes so that they can aggregate information across multiple clusters. 

This change adds that capability. An index within a rule can be defined as a single index string, a comma delimited string, or a YAML list. 

Two examples have been added to the example_rules folders.

### DIFF
--- a/elastalert/elastalert.py
+++ b/elastalert/elastalert.py
@@ -185,7 +185,7 @@ class ElastAlerter():
                         indexstring += ','
         else:
             indexstring = ",".join(indexes)
-            
+
         return indexstring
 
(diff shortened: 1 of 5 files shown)


NDevox:
+1 this would be a nice addition to complement chains available on watcher.

Qmando:
You could also just comma separate indices.
``index: logstash-foo-%Y.%m,logstash-bar-%Y.%m``
I'm not sure if this is necessary at all.

PR #309, opened 2015-11-13
IRC Alert - proper library added to setup

amoennin:


amoennin:
Going to fix this and create a new pull request later. 
The additions include an alerter that connects to an IRC channel, sends a message (alert), and then disconnects.

PR #311, opened 2015-11-16
Add VictorOps REST endpoint integration

aaronrea:
These commits add code and documentation for sending alerts via VictorOps (http://victorops.com).


aaronrea:
I'll test a little more locally.

PR #313, opened 2015-11-17
Adding class for victorops rest endpoint integration

aaronrea:


docs/source/ruletypes.rst
```diff
@@ -1088,6 +1088,21 @@ The alerter requires the following option:
...
+
+The alerter requires the following options:
+
+``victorops_api_key``: API key generated under the 'REST Endpoint' in the Integrations settings.
+
+``victorops_routing_key``: VictorOps routing key to route the alert to.
+
+Optional:
+
+``victorops_entity_display_name``: Humna-readable name of alerting entity. Used by VictorOps to correlate incidents by host througout the alert lifecycle.
```
Qmando: Should be ``victorops_entity_id``?

elastalert/alerts.py
```diff
@@ -624,3 +624,43 @@ def get_info(self):
...
+        for match in matches:
+            body += unicode(BasicMatchString(self.rule, match))
+            # Separate text of aggregated alerts with dashes
+            if len(matches) > 1:
+                body += '\n----------------------------------------\n'
+
+        # post to victorops
+        headers = {'content-type': 'application/json'}
+        payload = {
+            "message_type": self.victorops_message_type,
```
Qmando: Where does ``self.victorops_message_type`` come from?

elastalert/alerts.py
```diff
@@ -624,3 +624,43 @@ def get_info(self):
...
+
+class VictorOpsAlerter(Alerter):
+    """ Creates a VictorOps Incident for each alert """
+    required_options = frozenset(['victorops_api_key', 'victorops_routing_key'])
+
+    def __init__(self, rule):
+        super(VictorOpsAlerter, self).__init__(rule)
+        self.victorops_api_key = self.rule['victorops_api_key']
+        self.victorops_routing_key = self.rule['victorops_routing_key']
+        self.victorops_entity_id = self.rule['victorops_entity_id']
```
Qmando: If this is optional, it should be ``self.rule.get('victorops_entity_id')``.

I don't see it used anywhere right now.

aaronrea:
@Qmando thanks for the feedback. I think I botched a rebase -i, I'm somewhat new to git. I meant for victorops_entity_display_name to replace victorops_entity_id as the former is generally set by the user and the latter is derived from the machine. victorops_message_type actually has a set list of possible values which I am about to commit along with the rule.get change.

aaronrea:
@Qmando anything else needed to get this merged?

Qmando:
Merged as #351

PR #317, opened 2015-11-20
JIRA GroupPicker Custom Field Support

banjoey:
Added in support for custom fields in JIRA that are of Type GroupPicker. The parameters in the JSON is dependent upon the field type. Would eventually like to refactor to support all JIRA custom field types with a common syntax but don't have time right now.

### DIFF
--- a/elastalert/alerts.py
+++ b/elastalert/alerts.py
@@ -307,7 +307,7 @@ class JiraAlerter(Alerter):
         if self.assignee:
             self.jira_args['assignee'] = {'name': self.assignee}
         if self.custom_text_fields:
-            for key, value in self.custom_text_fields:
+            for key, value in self.custom_text_fields.items():
                 self.jira_args[key] = value
 
 
--- a/elastalert/alerts.py
+++ b/elastalert/alerts.py
@@ -277,7 +277,7 @@ class JiraAlerter(Alerter):
         self.get_account(self.rule['jira_account_file'])
         self.project = self.rule['jira_project']
         self.issue_type = self.rule['jira_issuetype']
-        self.security_level = self.rule['jira_security_level']
+        self.security_level = self.rule.get('jira_security_level')
         self.component = self.rule.get('jira_component')
         self.label = self.rule.get('jira_label')
         self.description = self.rule.get('jira_description', '')
@@ -301,7 +301,7 @@ class JiraAlerter(Alerter):
         self.jira_args = {'project': {'key': self.project},
                           'issuetype': {'name': self.issue_type}}
 
-        if self.security_level != "":
+        if self.security_level:
             self.jira_args['security'] = {'name': self.security_level}
 
         if self.component:
--- a/elastalert/alerts.py
+++ b/elastalert/alerts.py
@@ -277,6 +277,7 @@ class JiraAlerter(Alerter):
         self.get_account(self.rule['jira_account_file'])
         self.project = self.rule['jira_project']
         self.issue_type = self.rule['jira_issuetype']
+        self.security_level = self.rule['jira_security_level']
         self.component = self.rule.get('jira_component')
         self.label = self.rule.get('jira_label')
         self.description = self.rule.get('jira_description', '')
@@ -298,7 +299,8 @@ class JiraAlerter(Alerter):
             logging.warning(msg)
 
         self.jira_args = {'project': {'key': self.project},
-                          'issuetype': {'name': self.issue_type}}
+                          'issuetype': {'name': self.issue_type},
+                          'security': {'name': self.security_level}}
 
         if self.component:
             self.jira_args['components'] = [{'name': self.component}]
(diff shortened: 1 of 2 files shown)


elastalert/alerts.py
```diff
@@ -320,12 +322,22 @@ def __init__(self, rule):
...
+        if self.security_level:
+            self.jira_args['security'] = {'name': self.security_level}
+
         if self.component:
             self.jira_args['components'] = [{'name': self.component}]
         if self.label:
             self.jira_args['labels'] = [self.label]
         if self.assignee:
             self.jira_args['assignee'] = {'name': self.assignee}
+        # Note: because of the sometimes strange behavior of JIRA with custom field types, we have
```
danielpops: It shouldn't be a whole lot of work to hide the customfield_1234 exposure to the end-users.  There is a rest endpoint that JIRA exposes which gives all the custom field mappings:

`
>>> import requests
>>> r = requests.get('https://jira/rest/api/2/field')
>>> r.text
u'[{"id":"customfield_12013","name":"Report Type","custom":true,"orderable":true,"navigable":true,"searchable":true,"clauseNames":["cf[12013]","Report Type"],"schema":{"type":"string","custom":"com.atlassian.jira.plugin.system.customfieldtypes:select","customId":12013}}
,{"id":"fixVersions","name":"Fix Version/s","custom":false,"orderable":true,"navigable":true,"searchable":true,"clauseNames":["fixVersion"],"schema":{"type":"array","items":"version","system":"fixVersions"}},{"id":"customfield_11200","name":"Rank","custom":true,"orderab
le":true,"navigable":true,"searchable":true,"clauseNames":["cf[11200]","Rank"],"schema":{"type":"array","items":"string","custom":"com.pyxis.greenhopper.jira:gh-lexo-rank","customId":11200}},{"id":"customfield_12015","name":"SR/MTG #","custom":true,"orderable":true,"nav
igable":true,"searchable":true,"clauseNames":["cf[12015]","SR/MTG #"],"schema":{"type":"string","custom":"com.atlassian.jira.plugin.system.customfieldtypes:textfield","customId":12015}},{"id":"resolution","name":"Resolution","custom":false,"orderable":true,"navigable":t
rue,"searchable":true,"clauseNames":["resolution"],"schema":{"type":"resolution","system":"resolution"}},{"id":"customfield_11201","name":"Team Members","custom":true,"orderable":true,"navigable":true,"searchable":true,"clauseNames":["cf[11201]","Team Members"],"schema"
:{"type":"array","items":"user","custom":"com.atlassian.jira.plugin.system.customfieldtypes:multiuserpicker","customId":11201}},{"id":"customfield_12017","name":"Google Doc[...]`

Havingsaidthat, there's probably some edge cases and other things that you'd have to deal with to really pull it off correctly.  

This pull request has been outstanding for quite some time.  Have you been using this and has it been working for you, @banjoey ?

elastalert/alerts.py
```diff
@@ -320,12 +322,22 @@ def __init__(self, rule):
         self.jira_args = {'project': {'key': self.project},
                           'issuetype': {'name': self.issue_type}}
 
+        if self.security_level:
+            self.jira_args['security'] = {'name': self.security_level}
```
danielpops: If the field is actually called 'security' in the API, it might be better to expose it to the user as 'security' too, so that there is no ambiguity

elastalert/alerts.py
```diff
@@ -320,12 +322,22 @@ def __init__(self, rule):
...
             self.jira_args['labels'] = [self.label]
         if self.assignee:
             self.jira_args['assignee'] = {'name': self.assignee}
+        # Note: because of the sometimes strange behavior of JIRA with custom field types, we have
+        # to format this differently than a text field. Would be nice to allow the user to just
+        # specify the field name and field value and we take care of the rest, but that'd be a lot
+        # of work. For now, we support group picker fields only.
+        if self.custom_group_picker_fields:
+            for key, value in self.custom_group_picker_fields.items():
+                self.jira_args[key] = {'name': value}
```
danielpops: I tried to run with this implementation, but unfortunately it doesn't work for all scenarios.  A custom field (and even built-in fields) can be set in many different ways, depending on the schema of that field:

1. self.jira_args[key] = {'name': 'value'}
2. self.jira_args[key] = [{'name': 'value1'}, {'name': 'value2']
3. self.jira_args[key] = 'value'
4. self.jira_args[key] = ['value1', 'value2']
5. self.jira_args[key] = {'id': 'value'}
6. self.jira_args[key] = [{'id': 'value1'}, {'id': 'value2']
7. self.jira_args[key] = {'key': 'value'}
8. self.jira_args[key] = [{'key': 'value1'}, {'key': 'value2']

And so on potentially...  There needs to be a lot more smarts than blindly taking the custom fields and setting them in the format you've suggested.  Furthermore, it will be preferable if users can express their field names with the public facing names ('Software Users') rather than the internal representation (customfield_1234).

I have an idea for how to make this work for cases 1-4 above, && allow the user to either specify the customfield_1234 OR the friendly name in snake_case (e.g. 'Software Users' would become 'jira_software_users').  I will hope to send a new pull request within the next couple of days.

danielpops: FYI, see PR #558

PR #319, opened 2015-11-20
IRC Alert

amoennin:
These additions include an alerter that connects to an IRC channel, sends a message (alert), and then disconnects. The Python IRC library used here is an older version to accommodate both python 2.6 and 2.7. There is an example rule called "testIRCRule.yaml" which can be used to demonstrate a working alert, once configured. Thanks!

### DIFF
--- a/elastalert/ircalert.py
+++ b/elastalert/ircalert.py
@@ -29,7 +29,7 @@ class IRCAlerter(Alerter):
             self.password = 'None'
         self.botnick = 'alertbot'
         self.body = ''
-        self.reactor = irc.client.Reactor()
+        self.client = irc.client.IRC()
         wrapper = functools.partial(ssl.wrap_socket, ssl_version=ssl.PROTOCOL_TLSv1, ciphers="AES256-SHA")
         self.ssl_factory = irc.connection.Factory(wrapper=wrapper)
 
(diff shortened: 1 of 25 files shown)


elastalert/ircalert.py
```diff
@@ -0,0 +1,112 @@
...
+import irc
+import irc.client
+from alerts import Alerter
+from alerts import BasicMatchString
+
+
+class IRCAlerter(Alerter):
+    '''Connects to an IRC channel, sends alert, and leaves'''
+    required_options = frozenset(['irc_server', 'irc_port', 'irc_channel', 'irc_realname'])
+    getattr(logging, 'DEBUG')
```
Qmando: This doesn't do anything?

elastalert/ircalert.py
```diff
@@ -0,0 +1,112 @@
...
+    logging.basicConfig(level=logging.DEBUG, format='%(asctime)s %(levelname)-5s [%(name)s] %(message)s')
+    logger = logging.getLogger('IRCAlerter')
```
Qmando: It would probably be preferable to use
```
from elastalert.util import elastalert_logger
```

elastalert/ircalert.py
```diff
@@ -0,0 +1,112 @@
...
+            self.logger.debug("Added on_privmsg handler")
+            status = c.is_connected()
+            if status is True:
+                self.logger.info("Connected to IRC: %s" % status)
+                while not self.delivered:
+                    client.process_once(0.2)
+                client.disconnect_all("done")
+                client.process_once(0.2)
+                self.logger.info("Done")
+                time.sleep(1)
```
Qmando: Why sleep at the end? Is this to prevent spamming the channel when there are multiple alerts?

PR #320, opened 2015-11-23
Change query_key conversion to unicode

Qmando:
Fixes another spot where a UnicodeDecodeError sometimes occurs.


y-trobinso:
lgtm

PR #321, opened 2015-11-25
Change timeout to string in new_term rule for ES 2.0

Qmando:
People have reported that this fixes new_term rule in ES 2.0. I have confirmed that this does not break compatibility with ES < 2.0.


y-trobinso:
lgtm

PR #323, opened 2015-11-30
Better cardinality rule

zlosim:
Hi guys
for me your cardinality rule was unusable as you are not aggregating data via ES and in my case killing my cluster by requesting millions of events. Take a look and let me think what you think. Can polish it if you want to merge it. Suggestions are welcome


Qmando:
This makes a lot of sense to add and looks alright to me. Thanks a ton.

I think we can also keep the non aggregated version as well though. It enables a few things that you can't get (currently) with aggregations, such as using a query_key in addition to the cardinality field. Or using a compound cardinality field. Or we could add a feature to include one event from each different value for the field.

nsano-rururu:
Test are failing

nsano-rururu:

PR #324, opened 2015-12-02
added a proper timeout to the new_term urllib3 connection

Qmando:
The timeout in the search API request does nothing when urllib3 defaults to a 10 second timeout.


talyian:
lg2m

y-trobinso:
lgtm

PR #331, opened 2015-12-08
Add proxy support to slack alert

rafaelzacarias:
Add proxy support to slack alerter, by setting the option **slack_proxy**


Qmando:
Looks good. Could you add in a section into the docs as well?

rafaelzacarias:
@Qmando it's done :)

PR #332, opened 2015-12-08
find_recent_pending_alerts: ignore aggregates

jraby:
There's a few changes to `find_recent_pending_alerts` and `send_pending_alerts` in this PR:
 - Only fetch unsent alerts that do not have an `aggregate_id`.  (such entries were ignored in `send_pending_alerts` anyways)
 - force ordering of alerts by `alert_time`,  ascending. So we process earlier alerts first.
 - Stop skipping over entries that have an `aggregate_id` in `send_pending_alerts`. They are now filtered  by ES.
 - changed some comments and tweaked some log messages to help debugging.

All of this is to avoid some strange conditions when there are more than 1000 entries matching `alert_sent:false` when `find_recent_pending_alerts` is called.

Since `find_recent_pending_alerts` limits its search to 1000 entries and because it was using the default sort (`score` + `_id`), the search could return an incomplete and somewhat invalid set of result.
Some alerts could be forever delayed because the search would never return them and some alert could be incomplete since more matches might exist beyond the 1000 limit.

With this, I think most case are covered, unless you have more than 1000 rules.
I've left a comment regarding `max_aggregation`. If there's more entries than this limit, we're leaving cruft in ES.

There's another tweak that could be done when deleting sent alerts from ES,  I think it would be more efficient to do those delete a bulk call, but I'll test it and leave it for another PR.

### DIFF
--- a/elastalert/elastalert.py
+++ b/elastalert/elastalert.py
@@ -977,8 +977,7 @@ class ElastAlerter():
         query = {'query': {'query_string': {'query': '!_exists_:aggregate_id AND alert_sent:false'}},
                  'filter': {'range': {'alert_time': {'from': dt_to_ts(ts_now() - time_limit),
                                                      'to': dt_to_ts(ts_now())}}},
-                 'sort': { 'alert_time': {'order': 'asc'}},
-                }
+                 'sort': {'alert_time': {'order': 'asc'}}}
         if self.writeback_es:
             try:
                 res = self.writeback_es.search(index=self.writeback_index,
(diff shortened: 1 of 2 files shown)


jraby:
@Qmando what's your opinion on this?

#335 fixed the original issue I was experiencing, but I think it would be more efficient to let ES do the filtering of `aggregated_id` instead of doing it client side.

I'll look at fixing the test if you want to merge this.

Qmando:
This looks good. I didn't merge because of the failing Travis tests, which I haven't had the chance yet to investigate.

jraby:
right,   I'll have a look, I'm on a bugfix spree...

PR #336, opened 2015-12-09
Apply rule['max_query_size'] if defined

jraby:
The doc seems to indicate that this should be possible.


Qmando:
Looks good! Thanks for adding this. There have definitely been some inconsistencies between what is applied globally and what can be defined per rule.

Issue #326, opened 2015-12-03
search_phase_execution_exception errors


mattupstate:
I'm constantly getting `search_phase_execution_exception` errors and can't figure out why, even when I add the appropriate CLI flags to get more output.

```
$ elastalert --config /etc/elastalert/config.yaml --debug --es_debug --verbose --es_debug_trace /etc/elastalert/errors.log
ERROR:root:Error running query: TransportError(500, u'search_phase_execution_exception')
Skipping writing to ES: {'message': "Error running query: TransportError(500, u'search_phase_execution_exception')", 'traceback': ['Traceback (most recent call last):', '  File "/usr/local/lib/python2.7/site-packages/elastalert/elastalert.py", line 258, in get_hits', '    res = self.current_es.search(index=index, size=self.max_query_size, body=query, ignore_unavailable=True, **extra_args)', '  File "/usr/local/lib/python2.7/site-packages/elasticsearch/client/utils.py", line 69, in _wrapped', '    return func(*args, params=params, **kwargs)', '  File "/usr/local/lib/python2.7/site-packages/elasticsearch/client/__init__.py", line 530, in search', "    doc_type, '_search'), params=params, body=body)", '  File "/usr/local/lib/python2.7/site-packages/elasticsearch/transport.py", line 307, in perform_request', '    status, headers, data = connection.perform_request(method, url, params, body, ignore=ignore, timeout=timeout)', '  File "/usr/local/lib/python2.7/site-packages/elasticsearch/connection/http_urllib3.py", line 93, in perform_request', '    self._raise_error(response.status, raw_data)', '  File "/usr/local/lib/python2.7/site-packages/elasticsearch/connection/base.py", line 105, in _raise_error', '    raise HTTP_EXCEPTIONS.get(status_code, TransportError)(status_code, error_message, additional_info)', "TransportError: TransportError(500, u'search_phase_execution_exception')"], 'data': {'rule': 'CPU spike'}}
[...]

PR #338, opened 2015-12-09
Changed max_query_size, run_every and buffer_time defaults

Qmando:
See https://github.com/Yelp/elastalert/issues/326

ES > 2.1 will not accept 100,000 size for queries. This will lower the default to 10,000. I also lowered the example buffer_time to just 15 minutes so that more documents can be effectively queried. The original intent of buffer_time was to compensate for latency in indexing, which I think 15 minutes still suffices as a default.

I also lowered run_every in config.yaml.example to 1 minute. I think most people care about faster latency and can adjust down if it's hammering Elasticsearch too hard.

### DIFF
--- a/config.yaml.example
+++ b/config.yaml.example
@@ -5,12 +5,12 @@ rules_folder: example_rules
 # How often ElastAlert will query elasticsearch
 # The unit can be anything from weeks to seconds
 run_every:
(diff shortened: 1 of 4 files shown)


y-trobinso:
lgtm

PR #341, opened 2015-12-10
Pass alert_time to alerters via pipeline

jraby:
This allows the alerters to use `alert_time` to display it or to calculate the timerange which the current alert covers.

It doesn't seem very 'clean' to pass alert metadata via a parameter named `pipeline`.
In fact, I was about to add a `metadata` object to the alerters when I saw that `pipeline`  did exactly what I wanted.

From my point of view, the `pipeline` object's purpose is to pass alert metadata from one alerter to the next, and we might as well use it if we want to pass more info from ElastAlerter.
With that in mind, what would you think of renaming that parameter to `metadata` ?

### DIFF
--- a/elastalert/elastalert.py
+++ b/elastalert/elastalert.py
@@ -841,7 +841,7 @@ class ElastAlerter():
     def alert(self, matches, rule, alert_time=None):
         """ Wraps alerting, kibana linking and enhancements in an exception handler """
         try:
-            return self.send_alert(matches, rule, alert_time=None)
+            return self.send_alert(matches, rule, alert_time=alert_time)
         except Exception as e:
             self.handle_uncaught_exception(e, rule)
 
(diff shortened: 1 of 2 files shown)


elastalert/elastalert.py
```diff
@@ -841,7 +841,7 @@ def filters_from_kibana(self, rule, db_name):
     def alert(self, matches, rule, alert_time=None):
         """ Wraps alerting, kibana linking and enhancements in an exception handler """
         try:
-            return self.send_alert(matches, rule, alert_time=None)
```
Qmando: Oops. Nice catch.

Qmando:
I'm open to renaming. I think ``metadata`` makes sense. I created it just for passing JIRA issue IDs to other alerters, and I would love to see it used for more stuff.

jraby:
A colleague of mine wrote a templated email alerter using jinja and I wanted to put the alert time in the subject.  pipeline seemed like the way to feed that info into the alerters.

I'll see if I can rename pipeline to metadata (and keep pipeline for backward compat)

PR #343, opened 2015-12-10
rule['max_query_size'] tweaks

jraby:
Simply a refinement of #336 :

 - Update docs regarding per rule max_query_size
 - set the default value at rule config time instead of before every search


jraby:
Updated test config so the test finds the key.

However, I guess it would be better to test the inheritance of config values from global config to rule config in the tests instead of hardcoding values like this.

Issue #327, opened 2015-12-03
Slow searches - why not use filtered queries by default?


jraby:
I was trying to understand why the search were so slow for simple rules like this one:
```
name: HTTP 500
type: frequency
index: logstash-*
num_events: 1
timeframe:
  hours: 1
aggregation:
  hours: 1
realert:
  minutes: 0

filter:
- query:
    query_string:
      query: 'http_response_code:500'

[...]
```

For this rule, it seems that elastalert will generate elasticsearch queries similar to this:
```
{"filter": {"bool": {"must": [{"range": {"@timestamp": {"gt": "2015-12-02T22:40:29.757603Z", "lte": "2015-12-02T22:41:29.757603Z"}}}, {"query": {"query_string": {"query": "http_response_code:500"}}} ]}}, "sort": [{"@timestamp": {"order": "asc"}}]}
```

This is *really* slow,  it takes 15s+ to run on our machine.

Using a filtered query would be much more efficient, something like the following takes ~150ms to run here:
```
{"sort": [{"@timestamp": {"order": "asc"}}], "query": {"filtered": {"filter": {"bool": {"must": [{"range": {"@timestamp": {"gt": "2015-12-03T00:49:00Z", "lte": "2015-12-03T00:50:00Z"}}}, {"query": {"query_string": {"query": "http_response_code:400"}}}]}}}}}
```

From my perspective, it seems that elastalert should be using filtered queries by default for such rules. Am I missing something or is there a reason why it should not do so?

I'll hack up a proof of concept today.

Qmando:
I've not seen such drastic performance issues. Are you sure that the second query was not faster because the first had been cached? 

As I understand it, which I may be completely mistaken about, having a filtered query only gives you better performance if you have distinct filter and query sections. In your example, the query_string is still part of the bool: {must: [...]} and thus would not benefit from the other filters running first. If it were like
```
{
"query": {"query_string": {...} },
 "filter": {"bool": {...} }
}
```
then it could be optimized.

I am realizing now though that the top level filter is not cached by default. I think by changing it to be like
```
{ "filter":
    {"bool":
      {"must": [ ... ]
       "_cache": True
      }
   }
}
```
then the result of the query_string and all the other queries and filters would all be saved in a single cache.

I would definitely like to hear some more opinions on this and do some proper testing.

jraby:
Not sure about caching,  running both query twice gives the same results:
  - the original one is slow twice
  - the filtered query is always fast

[...]

PR #345, opened 2015-12-10
Replace match all + post filter by filtered query

jraby:
This is a proof of concept,  instead of using an implicit matchall query and applying a post filter, use a filtered query so that the filters get applied at query time.

See #327 for more background info. It give a major performance improvement in our case, but we're only using frequency rules at the moment, more tests would be welcome :)

(also, it breaks tests)

### DIFF
--- a/elastalert/elastalert.py
+++ b/elastalert/elastalert.py
@@ -190,10 +190,11 @@ class ElastAlerter():
         starttime = to_ts_func(starttime)
         endtime = to_ts_func(endtime)
         filters = copy.copy(filters)
-        query = {'filter': {'bool': {'must': filters}}}
+        es_filters = {'filter': {'bool': {'must': filters}}}
         if starttime and endtime:
-            query['filter']['bool']['must'].append({'range': {timestamp_field: {'gt': starttime,
+            es_filters['filter']['bool']['must'].insert(0,{'range': {timestamp_field: {'gt': starttime,
                                                                                 'lte': endtime}}})
+        query = {'query': {'filtered': es_filters}}
         if sort:
             query['sort'] = [{timestamp_field: {'order': 'desc' if desc else 'asc'}}]
         return query


elastalert/elastalert.py
```diff
@@ -190,10 +190,11 @@ def get_query(filters, starttime=None, endtime=None, sort=True, timestamp_field=
         starttime = to_ts_func(starttime)
         endtime = to_ts_func(endtime)
         filters = copy.copy(filters)
-        query = {'filter': {'bool': {'must': filters}}}
+        es_filters = {'filter': {'bool': {'must': filters}}}
         if starttime and endtime:
-            query['filter']['bool']['must'].append({'range': {timestamp_field: {'gt': starttime,
+            es_filters['filter']['bool']['must'].insert(0,{'range': {timestamp_field: {'gt': starttime,
```
Qmando: Why change append to insert?

jraby: so that the timerange filter is applied first: https://www.elastic.co/guide/en/elasticsearch/guide/current/_filter_order.html

I guess it only makes a difference when the query is running against a huge index.

Qmando: I wonder if it makes a difference that the time filter uses specific timestamps and therefore will not be cached. This seems very dependent on what the individual filters are, but I think either way will be acceptable.

Qmando:
Sorry this has taken me so long to get around to testing this. It does appear to perform better. I'll go ahead and fix the tests up and merge it.

jraby:
Ok great!

we've been running with this for months now and I didn't see any issue. But we're only using frequency rules.

sathishdsgithub:
@jraby 

Would you recommend me some options to increase query speed and get maximum query documents for every query? I use ES Version 5 and Graylog for log collection and parsing.

PR #353, opened 2015-12-17
fix #352 - missing values on spike alerts triggered by garbage_collect

### LINKED ISSUE #352: Spike rules and garbage_collect
I'm finding that many of my SpikeRule alerts are coming through with `<MISSING VALUE>` for fields I've included. After some debugging, I've narrowed it down to this:

1. With `alert_on_new_data: true`
2. A new `query_key` comes in at a high volume
3. The alert is triggered when `garbage_collect` occurs, by a placeholder value.
4. `handle_event` passes `self.cur_windows[qk].data[-1][0]` to `add_match`. This is the latest event, but it's the placeholder event, which doesn't have any of the `include:` fields that it should. The alert fires, but all the values I'm interested in are missing.

PR incoming for a proposed fix (to use the first event of the last instead of the current event count is 0). This is because I found that there was often more than one placeholder at the end of the current window.

ssandler:
Fixes #352 - input welcome on the best way to approach this.

### DIFF
--- a/elastalert/ruletypes.py
+++ b/elastalert/ruletypes.py
@@ -377,11 +377,10 @@ class SpikeRule(RuleType):
             self.ref_window_filled_once = True
 
         if self.find_matches(self.ref_windows[qk].count(), self.cur_windows[qk].count()):
-            # If placeholder (with 0 count) triggered alert, use a real event (the first one) instead of the latest (placeholder) event
-            if count == 0:
-                match = self.cur_windows[qk].data[0][0]
-            else:
-                match = self.cur_windows[qk].data[-1][0]
+            # skip over placeholder events which have count=0
+            for match, count in self.cur_windows[qk].data:
+                if count:
+                    break
 
             self.add_match(match, qk)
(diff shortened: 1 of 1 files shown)


elastalert/ruletypes.py
```diff
@@ -377,7 +377,12 @@ def handle_event(self, event, count, qk='all'):
             self.ref_window_filled_once = True
 
         if self.find_matches(self.ref_windows[qk].count(), self.cur_windows[qk].count()):
-            match = self.cur_windows[qk].data[-1][0]
+            # If placeholder (with 0 count) triggered alert, use a real event (the first one) instead of the latest (placeholder) event
+            if count == 0:
+                match = self.cur_windows[qk].data[0][0]
```
Qmando: I think technically data[0] could also be a placeholder event.
Perhaps better method might be something like
```
for match, count in self.cur_windows[qk].data:
    if count:
        break
```

ssandler:
Thanks @Qmando - tested with your version and it seemed to work. Updated the PR with that change.

PR #370, opened 2016-01-13
Added ability for required options to be split between inline and global

Qmando:
The requirements were previously checked against only the inline alert configuration, even though the they can be used at the top level.

For example, if you have something like
```
jira_server: jira.com
alert:
- jira:
    jira_project: "foo"
- jira:
    jira_project: "bar"
```
it would have complained about jira_server being missing, even though it can be shared between alerts.


y-trobinso:
lgtm

PR #378, opened 2016-01-20
Small optimisations

NDevox:
Making small changes to source code as I'm reading through it which I feel may be appropriate/useful or enhance the readability of the code.

This is a prospective pull - happy for it to be rejected as most changes are very minor.

### DIFF
--- a/elastalert/util.py
+++ b/elastalert/util.py
@@ -47,7 +47,7 @@ def lookup_es_key(lookup_dict, term):
         subkey = ''
 
         while subkeys:
-            subkey = subkeys.pop(0)
+            subkey += subkeys.pop(0)
             if subkey in go_deeper:
                 go_deeper = go_deeper[subkey]
                 subkey = ''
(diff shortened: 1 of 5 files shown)


NDevox:
Trying to figure out why this build is failing but not getting anywhere. Any ideas?

NDevox:
Excess lines apparently.

elastalert/util.py
```diff
@@ -47,8 +47,7 @@ def lookup_es_key(lookup_dict, term):
         subkey = ''
 
         while subkeys:
-            subkey += subkeys[0]
-            subkeys = subkeys[1:]
+            subkey = subkeys.pop(0)
```
Qmando: Should be ``subkey += subkeys.pop(0)``.
For example, if subkeys is ['foo', 'bar'], we want to test 'foo' then 'foo.bar'.

NDevox: @Qmando Good catch, fixed now.

fiunchinho:
Looks good to me

PR #380, opened 2016-01-21
Fixed duplicate alerts when using new_term

Qmando:
Fixes a bug when using new_term and use_terms_data where the same term would cause alerts to fire over and over again.

See equivalent code just 14 lines up in add_data.


talyian:
lg2m

y-trobinso:
lgtm

PR #389, opened 2016-02-01
Fixed bug where send_notify_email would crash on new rule errors

Qmando:
If an error occurs, new_rule will not exist. Instead, we can use the file name to help clarify the error message. All it previously used was the rule name, so the file name should also be sufficient.


ikoniaris:
:+1:  ship it

PR #411, opened 2016-02-26
scrolling support

omercnet:
Added support for scrolling, not sure this is the optimal way, please review

### DIFF
--- a/elastalert/elastalert.py
+++ b/elastalert/elastalert.py
@@ -262,7 +262,7 @@ class ElastAlerter():
 
         return processed_hits
 
-    def get_hits(self, rule, starttime, endtime, index):
+    def get_hits(self, rule, starttime, endtime, index, scroll=False):
         """ Query elasticsearch for the given rule and return the results.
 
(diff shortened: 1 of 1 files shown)


elastalert/elastalert.py
```diff
@@ -416,8 +429,10 @@ def run_query(self, rule, start=None, end=None):
                 rule_inst.add_data(data)
         # Warn if we hit max_query_size
         if self.num_hits - prev_num_hits == max_size and not rule.get('use_count_query'):
-            logging.warning("Hit max_query_size (%s) while querying for %s" % (max_size, rule['name']))
+            elastalert_logger.warning("Hit max_query_size (%s) while querying for %s, scrolling.." % (max_size, rule['name']))
```
omercnet: This warning is redundant when scrolling, could it be removed entirely ?

omercnet:
need to adjust tests to the changes, would be nice to have the scrolling search context timer as a config variable... to be continued

PR #439, opened 2016-03-22
Add alert_text_kw for keyword templating

tebeka:
This allows writing template with names and not positions which makes the template more readable.

New field `alert_text_kw` is a mapping of `match key` -> `template key`.


tebeka:
Thanks! Any schedule for the next release?

Qmando:
In the next few days.

tebeka:
Thanks

PR #446, opened 2016-03-25
Signed requests to AWS, so requests can be authenticated.

fiunchinho:
We are currently using AWS ElasticSearch and it's not possible to secure that service via network rules. The only way of sending secure requests to the service is by signing the requests with the authorization header, using the credentials of an AWS IAM user/role, with permissions on the AWS ElasticSearch service.

Since the authorization logic gets a little bit more complex, I've extracted that logic into its own class called `Auth`, that returns the right header depending on your credentials. This object support four different use cases:
- You don't want authorization header, so you don't pass `username`, `password` nor `aws_region`.
- You pass username and password to achieve basic authentication.
- You've deployed elastalert on an EC2 machine with an specific role, so you need to sign requests using that role's credentials. You just need to pass the `aws_region` argument.
- You need to sign requests and you are executing elastalert from your own machine, so you use a boto_profile passing the `boto_profile` argument.

If you guys think the code could be improved, just let me know!
Thank you.

### DIFF
--- a/docs/source/elastalert.rst
+++ b/docs/source/elastalert.rst
@@ -158,6 +158,10 @@ unless overwritten in the rule config. The default is "localhost".
 
 ``email_reply_to``: This sets the Reply-To header in emails. The default is the recipient address.
 
+``aws_region``: This makes ElastAlert to sign HTTP requests when using Amazon ElasticSearch Service. It'll use instance role keys to sign the requests.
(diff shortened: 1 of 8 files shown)


Qmando:
Looks good. Only thing I would like to see is a quick mention in the documentation. :)

PR #447, opened 2016-03-25
Added FAQ

Qmando:
Added a FAQ to the readme. Tested format with https://jbt.github.io/markdown-editor

I would love suggestions as to more questions to add.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,160 @@ of each other.
 - ``example_new_term.yaml`` is an example of the "new term" rule type, which alerts when a new value appears in a field or fields. In this example,
 an email is sent when a new value of ("username", "computer") is encountered in example login logs.
 
+## Frequently Asked Questions
+
+### My rule is not getting any hits?
+
+So you've managed to set up ElastAlert, write a rule, and run it, but nothing happens, or it says ``0 query hits``. First of all, I reccommend using the command ``elastalert-test-rule rule.yaml`` to debug. It will show you how many documents match your filters for the last 24 hours (or more, see ``--help``), and then shows you if any alerts would have fired. If you have a filter in your rule, remove it and try again. This will show you if the index is correct and that you have at least some documents. If you have a filter in Kibana and want to recreate it in ElastAlert, you probably want to use a query string. Your filter will look like
+
+```
+filter:
+- query:
+    query_string:
+      query: "foo: bar AND baz: abc*"
+```
+If you receive an error that Elasticsearch is unable to parse it, it's likely the YAML is not spaced correctly, and the filter is not in the right format. If you are using other types of filters, like ``term``, and common pitfall is not realizing that you may need to use the analyzed token. This is the default if you are using Logstash. For example,
+
+```
+filter:
+- term:
+    foo: "Test Document"
+```
+
+will not match even if the original value for ``foo`` was exactly "Test Document". Instead, you want to use ``foo.raw``. If you are still having trouble troubleshooting why your documents do not match, try running ElastAlert with ``--es_debug_trace /path/to/file.log``. This will log the queries made to Elasticsearch in full so that you can see exactly what is happening.
+
+### I got hits, why didn't I get an alert?
+
+If you got logs that had ``X query hits, 0 matches, 0 alerts sent``, it depends on the ``type`` why you didn't get any alerts. If ``type: any``, a match will occur for every hit. If you are using ``type: frequency``, ``num_events`` must occur within ``timeframe`` of each other for a match to occur. Different rules apply for different rule types.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -106,6 +106,160 @@ of each other.
 - ``example_new_term.yaml`` is an example of the "new term" rule type, which alerts when a new value appears in a field or fields. In this example,
 an email is sent when a new value of ("username", "computer") is encountered in example login logs.
 
+## Frequently Asked Questions
+
+### My rule is not getting any hits?
+
+So you've managed to set up ElastAlert, write a rule, and run it, but nothing happens, or it says ``0 query hits``. First of all, I reccommend using the command ``elastalert-test-rule rule.yaml`` to debug. It will show you how many documents match your filters for the last 24 hours (or more, see ``--help``), and then shows you if any alerts would have fired. If you have a filter in your rule, remove it and try again. This will show you if the index is correct and that you have at least some documents. If you have a filter in Kibana and want to recreate it in ElastAlert, you probably want to use a query string. Your filter will look like
```
ikoniaris: "I" --> "we" perhaps.

README.md
```diff
@@ -106,6 +106,160 @@ of each other.
...
+
+```
+filter:
+- query:
+    query_string:
+      query: "foo: bar AND baz: abc*"
+```
+If you receive an error that Elasticsearch is unable to parse it, it's likely the YAML is not spaced correctly, and the filter is not in the right format. If you are using other types of filters, like ``term``, and common pitfall is not realizing that you may need to use the analyzed token. This is the default if you are using Logstash. For example,
```
ikoniaris: 1. "and common" --> "a common"
2. I think you actually meant to say the not_analyzed token instead.

README.md
```diff
@@ -106,6 +106,160 @@ of each other.
...
+
+If you see ``X matches, 0 alerts sent``, this may occur for several reasons. If you set ``aggregation``, the alert will not be sent until after that time has elapsed. If you have gotten an alert for this same rule before, that rule may be silenced for a period of time. The default is one minute between alerts. If a rule is silenced, you will see ``Ignoring match for silenced rule`` in the logs.
+
+If you see ``X alerts sent`` but didn't get any alert, it's probably related to the alert configuration. If you are using the ``--debug`` flag, you will not receive any alerts. Instead, the alert text will be written to the console. Use ``--verbose`` to achieve the same affects without preventing alerts. If you are using email alert, make sure you have it configured for an SMTP server. By default, it will connect to localhost on port 25. It will also use the word "elastalert" as the from address. Some SMTP servers will rejects this because it does not have a domain while others will add their own domain automatically. See the email section in the documentation for how to configure this.
```
ikoniaris: 1. "from" address --> "From:" address
2. "rejects" --> "reject"

README.md
```diff
@@ -106,6 +106,160 @@ of each other.
...
+
+You can also prevent duplicates based on a certain field by using ``query_key``. For example, to prevent multiple alerts for the same user, you might use
+
+```
+realert:
+  hours: 8
+query_key: user
+```
+
+Note that this will also affect how many rule types work. If you are using ``type: frequency`` for example, ``num_events`` for a single value of ``query_key`` must occur before an alert will be sent. You can also use a compound of multiple fields for this key. For example, if you only wanted to receieve an alert once for a specific error and hostname, you could use
```
ikoniaris: "how many rule types work" is a bit misleading, like meaning some types might stop working. Change it to something like "the way many rule types work".

README.md
```diff
@@ -106,6 +106,160 @@ of each other.
...
+```
+
+Another thing you could change is ``buffer_time``. By default, ElastAlert will query large overlapping windows in order to ensure that it does not miss any events, even if they are indexed in real time. In config.yaml, you can adjust ``buffer_time`` to a smaller number to only query the most recent few minutes.
+
+```
+buffer_time:
+  minutes: 5
+```
+
+By default, ElastAlert will download every document in full before making processing them. Instead, you can have ElastAlert simply get a count of the number of documents that have occured in between each query. To do this, set ``use_count_query: true``. This cannot be used if you use ``query_key``, because ElastAlert will not know the contents of each documents, just the total number of them. This also reduces the precision of alerts, because all documents that occur between each query will be rounded to a single timestamp.
```
ikoniaris: 1. "before making processing" --> "before processing"
2. "all documents that occur" --> "all events that occur"

README.md
```diff
@@ -106,6 +106,160 @@ of each other.
...
+
+### Can I perform aggregations?
+
+The only aggregation supported currently is a terms aggregation, by setting ``use_terms_query``.
+
+### I'm not using @timestamp, what do I do?
+
+You can use ``timestamp_field`` to change which field ElastAlert will use as the timestamp. You can use ``timestamp_type`` to change it between ISO 8601 and unix timestamps. You must have some kind of timestamp for ElastAlert to work. If your events are not in real time, you can use ``query_delay`` and ``buffer_time`` to adjust when ElastAlert will look for documents.
+
+### I'm using flatline but I don't any alerts
```
ikoniaris: "I don't any" --> "I don't see any"?

ikoniaris:
@Qmando I added some comments, see above please.

Issue #450, opened 2016-03-28
Exception in `get_top_counts` prevents alert from being sent


lchi:
I'm using elastalert to query for HTTP 500 error codes and I'm running into an issue where an uncaught exception in `get_top_counts` is causing ElastAlert to not send the alert that was triggered (stacktrace below). Is this intentional? Currently this is a problem because we are seeing timeouts on the top counts queries and as a result not being alerted to a problem.

I can set `disable_rules_on_error` to `False` and set `notify_email` but IMO an alert should still be sent but without the alert text filled in.

I'm happy to submit a patch if the intended behavior is to send an alert even if `get_top_counts` fails for some reason.

```
WARNING:elasticsearch:GET http://<ip>:9200/logstash-*/<index>/_search?search_type=count&ignore_unavailable=true [status:N/A request:10.010s]
Traceback (most recent call last):
  File "/usr/lib/python2.6/site-packages/elasticsearch/connection/http_urllib3.py", line 78, in perform_request
    response = self.pool.urlopen(method, url, body, retries=False, headers=self.headers, **kw)
  File "/usr/lib/python2.6/site-packages/urllib3/connectionpool.py", line 597, in urlopen
    _stacktrace=sys.exc_info()[2])
  File "/usr/lib/python2.6/site-packages/urllib3/util/retry.py", line 222, in increment
    raise six.reraise(type(error), error, _stacktrace)
  File "/usr/lib/python2.6/site-packages/urllib3/connectionpool.py", line 544, in urlopen
    body=body, headers=headers)
  File "/usr/lib/python2.6/site-packages/urllib3/connectionpool.py", line 376, in _make_request
    self._raise_timeout(err=e, url=url, timeout_value=read_timeout)
  File "/usr/lib/python2.6/site-packages/urllib3/connectionpool.py", line 304, in _raise_timeout
    raise ReadTimeoutError(self, url, "Read timed out. (read timeout=%s)" % timeout_value)
ReadTimeoutError: HTTPConnectionPool(host='<ip>', port=9200): Read timed out. (read timeout=10)
ERROR:root:Error running query: ConnectionTimeout caused by - ReadTimeoutError(HTTPConnectionPool(host='<ip>', port=9200): Read timed out. (read timeout=10))
ERROR:root:Traceback (most recent call last):
  File "/home/.../elastalert/elastalert/elastalert.py", line 867, in alert
    return self.send_alert(matches, rule, alert_time=alert_time)
[...]

PR #452, opened 2016-03-29
Check for NoneType hits_terms

lchi:
PR to address #450. This should prevent the `NoneType` error that occurs when `get_hits_terms` returns `None`. I also chose to edit the top counts text that gets displayed so that it doesn't appear empty.

Let me know if there's anything that should be changed, and thanks for this tool!


Qmando:
Looks. Thanks a ton!

PR #454, opened 2016-03-29
Filtered queries

Qmando:
This improves the speed of queries by a significant amount. Original PR is https://github.com/Yelp/elastalert/pull/345. This fixes some tests, as well as use_terms_query and use_count_query.

Manually tested use_terms_query and use_count_query, verified they work. Tested on a set of rules over a specific period and it performed roughly twice as fast.


ikoniaris:
Looks good to me, but please be aware that according to Elastic this is to be deprecated in ES 2.0

jraby:
Oh indeed: https://www.elastic.co/blog/better-query-execution-coming-elasticsearch-2-0

I'm not sure the new syntax works with ES<2.0 however.

PR #458, opened 2016-03-31
Set `es_conn_timeout` in rule conf

lchi:
The rule configuration currently does not inherit the base config's setting for `es_conn_timeout` so the `Elasticsearch` instance created in each `run_rule` call via `new_elasticsearch` doesn't use the timeout specified in the config.

This was causing an issue where I had set `es_conn_timeout` in `config.yaml`, used that config using the `--config` flag but ended up with a 10 second timeout.


Qmando:
Thanks. Eventually I need to but a better mechanism for getting defaults from config.yaml instead of cherry picking each individual one.

PR #486, opened 2016-04-20
convert slack_webhook_url to array in order to allow alerting to mult…

msil:
…iple slack channels


Qmando:
Ideally this could be a single string OR a list of strings. We've got several other options of type "arrayOfString" which allow this behavior, like ``email`` for example.

Qmando:
You can also have multiple slack alerters attached to a single rule.

```
alert:
 - slack:
     slack_webhook_url: https://abc.slack.com
 - slack:
     slack_webhook_url: https://foo.slack.com
```

Although, I will probably accept this as well, just so people can do it both ways. I don't think maybe people realize that the above is possible.

PR #498, opened 2016-04-28
Allow to use custom timestamp format to fix #443

### LINKED ISSUE #443: custom date format
i've had a read through the docs and couldnt find anything, but i was wondering if its possible to use a custom date format somewhere in the configuration. the mapping for my date field is defined as 'date_hour_minute_second', and when running elastalert-test-rule i get an error:

ElasticsearchParseException[failed to parse date field [2016-03-23T09:22:47.536136Z], tried both date format [date_hour_minute_second], and timestamp number]; nested: IllegalArgumentException[Invalid format: "2016-03-23T09:22:47.536136Z" is malformed at ".536136Z"]; }

unicell:


elastalert/config.py
```diff
@@ -158,6 +160,11 @@ def load_options(rule, conf, args=None):
     elif rule['timestamp_type'] == 'unix_ms':
         rule['ts_to_dt'] = unixms_to_dt
         rule['dt_to_ts'] = dt_to_unixms
+    elif rule['timestamp_type'] == 'custom':
+        def _dt_to_ts_with_format(dt):
+            return dt_to_ts_with_format(dt, ts_format=rule['timestamp_format'])
+        rule['ts_to_dt'] = ts_to_dt
```
Qmando: Do you need to create a custom ts_to_dt function? Otherwise, it's just using ``dateutil.parser.parse(timestamp)``, which may work for some formats, but not necessarily others. It could probably be ``datetime.datetime.strptime(ts, format)`` and then have the UTC timezone added by default like the normal ts_to_dt function does.

unicell: done

unicell:
oops, just noticed it failed the CI.

unicell:
@Qmando could you take a look? I hope I addressed your comment in the initial PR request.

PR #499, opened 2016-04-29
fixed issue of converting elasticsearch type date into float

vaibhavkhanduja:
While fetching records with timestamp, declared as type "date" in elasticsearch, it is required to convert the field to float or integer. This is needed since return type comes as unicode string.


Qmando:
Thanks. I've included this fix in #506

Issue #505, opened 2016-05-02
Support Nested Time Fields


zinic:
Currently, elastalert allows a user to specify a custom timestamp field in a rule configuration YAML file.

Example rule:
```yaml
name: Example Rule
type: frequency
timestamp_field: P.time

...
```

However, the code in [process_hits](https://github.com/Yelp/elastalert/blob/master/elastalert/elastalert.py#L266) given this rule configuration fails with the following exception:
```
ERROR:root:Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/elastalert/elastalert.py", line 748, in run_all_rules
    num_matches = self.run_rule(rule, endtime, self.starttime)
  File "/usr/local/lib/python2.7/dist-packages/elastalert/elastalert.py", line 541, in run_rule
    if not self.run_query(rule, rule['starttime'], endtime):
  File "/usr/local/lib/python2.7/dist-packages/elastalert/elastalert.py", line 421, in run_query
    data = self.get_hits(rule, start, end, index)
  File "/usr/local/lib/python2.7/dist-packages/elastalert/elastalert.py", line 312, in get_hits
    hits = self.process_hits(rule, hits)
  File "/usr/local/lib/python2.7/dist-packages/elastalert/elastalert.py", line 266, in process_hits
    hit['_source'][rule['timestamp_field']] = rule['ts_to_dt'](hit['_source'][rule['timestamp_field']])
KeyError: 'P.time'
```

Are there any plans to better support nested timestamp fields?

Qmando:
This should definitely be supported. Anywhere where timestamp_field is referenced should be replaced with a lookup_es_key from util.py, which will automatically resolve nested fields.

zinic:
Awesome, I'll code up a PR for this then if that's okay. I could really use the functionality and it appears that the tools are already in place - just gotta use them here.

PR #507, opened 2016-05-03
Add support for nested timestamp fields

zinic:
Related issue: https://github.com/Yelp/elastalert/issues/505

This PR adds the following functionality:
* A new function in `util.py` named `set_es_key` which allows a user to set the value of an ES key regardless of nesting or other ES caveats outlined in `lookup_es_key`.
* A new test to outline expectations in `lookup_es_key` and `set_es_key` as well as validate them.
* Minor changes to `elasticsearch.py` and `ruletypes.py` to use `lookup_es_key` and `set_es_key` where appropriate in regards to utilization of nested timestamp fields.

**Notes**
I modified where some of the comments live in `util.py` by adding a third function that both `lookup_es_key` and `set_es_key` delegate to for the search. I wasn't sure where this comment should live since it's rather large - if this feels incorrect please let me know.

I have also changed some of the names in the search function to what I think better communicates intent. However, if this is not the case, I am more than happy to update the PR to have them changed back!

### DIFF
--- a/elastalert/elastalert.py
+++ b/elastalert/elastalert.py
@@ -33,6 +33,7 @@ from util import EAException
 from util import elastalert_logger
 from util import format_index
 from util import lookup_es_key
+from util import set_es_key
 from util import pretty_ts
 from util import seconds
 from util import ts_add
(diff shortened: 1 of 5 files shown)


Qmando:
Looks good. Thanks for doing this work.

PR #537, opened 2016-05-20
Search for pending aggergate alerts before starting new one

Qmando:
Fixes a bug where if an aggregation has been started, restarting elastalert would always start a new aggregation.

Now, before when a new aggregation starts, we will query for a future alert event that hasn't been set. If we find one, use it's alert_time and _id for the aggregation.

Updated existing tests but did not add any tests for this case specifically. I've tested this manually by starting an aggregation then restarting elastalert and verifying subsequent alerts were added to the original.


y-trobinso:
lgtm, after conversation off-thread

Issue #54, opened 2015-04-03
Multiple JIRA labels and components


Qmando:
You can only add a single label and component currently. You should be able to add more than one.

0x9090:
+1
This would greatly help organize our Jira workflow

PR #558, opened 2016-06-03
Dpopes alert metadata and more jira fields

danielpops:
@Qmando Here is a pull request for adding additional alert metadata (owner, priority) && support for alerts to refer and set arbitrary JIRA fields (and also addressing #54 while I was at it).  I have done some manual testing and it appears to work as advertised, but I will be adding more automated tests right now.  Just wanted to get the initial stab out there so that you can take a look if you're so inclined.

### DIFF
--- a/elastalert/alerts.py
+++ b/elastalert/alerts.py
@@ -313,6 +313,7 @@ class JiraAlerter(Alerter):
         self.bump_tickets = self.rule.get('jira_bump_tickets', False)
         self.bump_not_in_statuses = self.rule.get('jira_bump_not_in_statuses')
         self.bump_in_statuses = self.rule.get('jira_bump_in_statuses')
+        self.watchers = self.rule.get('jira_watchers')
 
         if self.bump_in_statuses and self.bump_not_in_statuses:
             msg = 'Both jira_bump_in_statuses (%s) and jira_bump_not_in_statuses (%s) are set.' % \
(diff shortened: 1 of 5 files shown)


elastalert/alerts.py
```diff
@@ -340,6 +395,71 @@ def __init__(self, rule):
...
+                            self.jira_args[arg_name] = [{'name': value}]
+                        else:
+                            self.jira_args[arg_name] = [{'name': v} for v in value]
+                # Handle non-array types
+                else:
+                    # String type
+                    if arg_type == 'string':
+                        self.jira_args[arg_name] = value
+                    # Number type
+                    elif arg_type == 'number':
```
danielpops: FYI, here are all the types that I see defined in our JIRA server:

```
>>> print '\n'.join(sorted(set([b['schema']['type'] for b in j.fields() if 'schema' in b and b['schema']['type'] != 'array'])))
date
datetime
issuetype
number
priority
progress
project
resolution
securitylevel
status
string
timetracking
user
version
```



date and datetime can be passed as simple string values, so long as they parse to dates or datetimes
string can obviously be passed as a simple string value
number can be passed as a simple value, but needs to be an int (and not a string)
securitylevel can be passed with "name": _value_, so long as _value_ is in the supported enum of securitylevel strings in the jira server
user can be passed with "name": _value_, so long as _value_ is a valid user in the jira server

i haven't found a use case for version, timetracking, or progress.  That doesn't mean that the use cases don't exist, I just either wasn't able to surface the type on my test JIRA project, or I couldn't find a meaningful way to set the value.  Either way, I don't think that it should be considered a blocker if we're not able to anticipate and support every single possible type of field for now.  As the need arises, support can be added.

PR #567, opened 2016-06-10
Set up travis to auto-publish to pypi

danielpops:
This was stolen from [threat intel](https://github.com/Yelp/threat_intel/blob/master/.travis.yml)


jjsendor:
@danielpops I don't think it works there. Was talking with @tomelm to fix it someday.

PR #578, opened 2016-06-15
Split new_term queries into steps to avoid crazy timeouts

Qmando:
When calling get_all_terms when a new_term rule starts, split the query (which is default 30 days) into (default 1 day) chunks. This can avoid timeouts because this can be a very expensive query to run. The step size is controllable via ``window_step_size``.

### DIFF
--- a/elastalert/ruletypes.py
+++ b/elastalert/ruletypes.py
@@ -583,7 +583,7 @@ class NewTermsRule(RuleType):
                 if tmp_start == tmp_end:
                     break
                 tmp_start = tmp_end
-                tmp_end = min(end, tmp_end + step)
+                tmp_end = min(tmp_start + step, end)
                 time_filter[self.rules['timestamp_field']] = {'lt': dt_to_ts(tmp_end), 'gte': dt_to_ts(tmp_start)}
 
             for key, values in self.seen_values.iteritems():
@@ -600,9 +600,7 @@ class NewTermsRule(RuleType):
                         elastalert_logger.info('Found no values for %s' % (field))
                     continue
                 self.seen_values[key] = list(set(values))
-                if type(key) == str:
-                    # Only print this number out for single fields
-                    elastalert_logger.info('Found %s unique values for %s' % (len(values), key))
+                elastalert_logger.info('Found %s unique values for %s' % (len(values), key))
 
     def flatten_aggregation_hierarchy(self, root, hierarchy_tuple=()):
         """ For nested aggregations, the results come back in the following format:
--- a/elastalert/ruletypes.py
+++ b/elastalert/ruletypes.py
@@ -4,17 +4,17 @@ import datetime
 
 from blist import sortedlist
 from elasticsearch.client import Elasticsearch
+from util import add_raw_postfix
 from util import dt_to_ts
 from util import EAException
 from util import elastalert_logger
 from util import format_index
 from util import hashable
 from util import lookup_es_key
+from util import new_get_event_ts
 from util import pretty_ts
 from util import ts_now
 from util import ts_to_dt
-from util import new_get_event_ts
-from util import add_raw_postfix
 
 
 class RuleType(object):
@@ -517,7 +517,7 @@ class NewTermsRule(RuleType):
             self.get_all_terms(args)
         except Exception as e:
             # Refuse to start if we cannot get existing terms
-            raise EAException('Error searching for existing terms: %s' % (e))
+            raise EAException('Error searching for existing terms: %s' % (repr(e)))
 
     def get_all_terms(self, args):
         """ Performs a terms aggregation for each field to get every existing term. """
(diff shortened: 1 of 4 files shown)


elastalert/ruletypes.py
```diff
@@ -555,33 +556,53 @@ def get_all_terms(self, args):
...
+            # Query the entire time range in small chunks
+            while tmp_start < end:
+                if self.rules.get('use_strftime_index'):
+                    index = format_index(self.rules['index'], tmp_start, tmp_end)
+                else:
+                    index = self.rules['index']
+                res = self.es.search(body=query, index=index, ignore_unavailable=True, timeout='50s')
+                if 'aggregations' in res:
+                    buckets = res['aggregations']['filtered']['values']['buckets']
+                    if type(field) == list:
```
y-trobinso: Note that there are other collections than type(..) == list, and that this will not support subclassing whatsoever. Consider isinstance(..) or issubclass(..), and checking for __iter__ and other dunder methods you intend to use, if you care about that.

elastalert/ruletypes.py
```diff
@@ -555,33 +556,53 @@ def get_all_terms(self, args):
...
-                    self.seen_values[field] = keys
-                    elastalert_logger.info('Found %s unique values for %s' % (len(keys), field))
-            else:
-                self.seen_values[field] = []
-                elastalert_logger.info('Found no values for %s' % (field))
+                    else:
+                        elastalert_logger.info('Found no values for %s' % (field))
+                    continue
+                self.seen_values[key] = list(set(values))
+                if type(key) == str:
```
y-trobinso: Same considerations as above, different dunder methods.

tests/rules_test.py
```diff
@@ -474,9 +475,32 @@ def test_new_term():
     with mock.patch('elastalert.ruletypes.Elasticsearch') as mock_es:
         mock_es.return_value = mock.Mock()
         mock_es.return_value.search.return_value = mock_res
+        call_args = []
+
+        # search is called with a mutable dict containing timestamps, this is required to test
```
y-trobinso: A better phrasing would be to call this a callback that you've injected into your mock. But lgtm.

tests/rules_test.py
```diff
@@ -474,9 +475,32 @@ def test_new_term():
...
 
-        assert rule.es.search.call_count == 2
+    # 30 day default range, 1 day default step, times 2 fields
+    assert rule.es.search.call_count == 60
+
+    # Assert that all calls have the proper ordering of time ranges
+    old_ts = '2010-01-01T00:00:00Z'
+    old_field = ''
+    for call in call_args:
+        field = call[1]['body']['aggs']['filtered']['aggs']['values']['terms']['field']
```
y-trobinso: These calls make Demeter sad: https://en.wikipedia.org/wiki/Law_of_Demeter

However, as always, we've found no better way to do this. Consider breaking these calls up into individual classes if you care to clean these up and make them easier to understand.

y-trobinso:
lgtm. I've added some open issues, but none of these should block shipping this change.

elastalert/ruletypes.py
```diff
@@ -535,17 +535,18 @@ def get_all_terms(self, args):
...
             end = ts_now()
         start = end - window_size
-        if self.rules.get('use_strftime_index'):
-            index = format_index(self.rules['index'], start, end)
-        else:
-            index = self.rules['index']
-        time_filter = {self.rules['timestamp_field']: {'lte': dt_to_ts(end), 'gte': dt_to_ts(start)}}
-        query_template['filter'] = {'bool': {'must': [{'range': time_filter}]}}
-        query = {'aggs': {'filtered': query_template}}
+        step = datetime.timedelta(**self.rules.get('window_step_size', {'days': 1}))
```
danielpops: Since you're adding this new rule property, you should update the documentation too!

elastalert/ruletypes.py
danielpops: Those are actually my fault, from a previous set of changes I made to support composite keys.  I think you're right that isinstance is probably a better check.  However, do note that these are coming from yaml, and I'm not sure if the user-provided fields will ever take the form of a collection that's not a list (at least in this context, particularly given the schema validation that we enforce on the rule files themselves)

elastalert/ruletypes.py
```diff
@@ -555,33 +556,53 @@ def get_all_terms(self, args):
...
+                            self.seen_values[tuple(field)] += self.flatten_aggregation_hierarchy(bucket)
+                    else:
+                        keys = [bucket['key'] for bucket in buckets]
+                        self.seen_values[field] += keys
+                else:
+                    self.seen_values.setdefault(field, [])
+                if tmp_start == tmp_end:
+                    break
+                tmp_start = tmp_end
+                tmp_end = min(end, tmp_end + step)
```
danielpops: nit: even though tmp_start and tmp_end are the same thing at this point, I feel like it reads better as:
`tmp_end = min(tmp_start + step, end)`

elastalert/ruletypes.py
```diff
@@ -555,33 +556,53 @@ def get_all_terms(self, args):
...
-                    elastalert_logger.info('Found %s unique values for %s' % (len(keys), field))
-            else:
-                self.seen_values[field] = []
-                elastalert_logger.info('Found no values for %s' % (field))
+                    else:
+                        elastalert_logger.info('Found no values for %s' % (field))
+                    continue
+                self.seen_values[key] = list(set(values))
+                if type(key) == str:
+                    # Only print this number out for single fields
```
danielpops: Why not print out the number for composite fields too?

elastalert/ruletypes.py
```diff
@@ -535,17 +535,18 @@ def get_all_terms(self, args):
...
-        time_filter = {self.rules['timestamp_field']: {'lte': dt_to_ts(end), 'gte': dt_to_ts(start)}}
-        query_template['filter'] = {'bool': {'must': [{'range': time_filter}]}}
-        query = {'aggs': {'filtered': query_template}}
+        step = datetime.timedelta(**self.rules.get('window_step_size', {'days': 1}))
 
         for field in self.fields:
+            tmp_start = start
+            tmp_end = min(start + step, end)
+
+            time_filter = {self.rules['timestamp_field']: {'lt': dt_to_ts(tmp_end), 'gte': dt_to_ts(tmp_start)}}
```
danielpops: Before introducing the step, the time range was fully inclusive (gte && lte).  With the introduction of the step, the full range you're querying is now inclusive of the start, but exclusive of the end.  It's probably a minor detail, but it's probably worth either addressing in code, or calling out in the docs!

danielpops:
LGTM.  Just a few minor comments.

elastalert/ruletypes.py
Qmando: I think people expect inclusive on both ends more often than not, though there is no real consistency throughout elastalert unfortunately.

danielpops:
Shippppppppppppit! (then tagggggggggggggg it, and let's see Travis do his pypi magic :))

PR #583, opened 2016-06-15
Added run_enhancements_first option

Qmando:
Adds a new option, ``run_enhancements_first``, which lets you run ehancements before they are added to an aggregation and before they add a ``realert`` setting. This is useful if your alerts might be dropped and you don't want that to silence other alerts, or if you need them to run before an aggregation gets sent.


danielpops:
Ship it!

Issue #592, opened 2016-06-21
Problems with "command" alert


MaxFlanders:
Hi there, I'm encountering all sorts of strange issues trying to set up a command alert. Essentially all I want to do is hit an http endpoint with any matching documents. This is the structure I've got set up for my command alert:

```
    alert:
     - command:
         command: "curl -XPOST -H \"Content-Type: application/json\" \"https://my.endpoint\" -d@-"
    pipe_match_json: true
```

The -d@- option causes curl commands to read from stdin, which as I understand, pipe_match_json should provide.

When I do I get this strange error though:
    `elastalert_error - {'message': 'Error while running alert command: Error while running command c u r l   - X P O S T   - H   " C o n t e n t - T y p e :   a p p l i c a t i o n / j s o n "   " h t t p s : / / m y . e n d p o i n t"   - d @ -: [Errno 2] No such file or directory', 'traceback': ['Traceback (most recent call last):', '  File "/usr/local/lib/python2.7/dist-packages/elastalert-0.0.84-py2.7.egg/elastalert/elastalert.py", line 976, in send_alert', '    alert.alert(matches)', '  File "/usr/local/lib/python2.7/dist-packages/elastalert-0.0.84-py2.7.egg/elastalert/alerts.py", line 660, in alert', '    raise EAException("Error while running command %s: %s" % (\' \'.join(command), e))', 'EAException: Error while running command c u r l   - X P O S T   - H   " C o n t e n t - T y p e :   a p p l i c a t i o n / j s o n "   " h t t p s : / / m y . e n d p o i n t "   - d @ -: [Errno 2] No such file or directory'], 'data': {'rule': 'Google Microservice Error Alert'}}`

In which it seems to be putting spaces in between every letter of my argument.
Any thought on what is going on here??

Thanks!

326TimesBetter:
Sorry I'm still learning the syntax to post code in these forums, those backticks around the alert definitely are not in the rule.yaml file lol

Qmando:
try
```
alert:
- command:
    command: ["curl -XPOST -H \"Content-Type: application/json\" \"https://my.endpoint\" -d@-"]
pipe_match_json: true 
```

This needs to be fixed actually. It interprets a string as a list in this case.

Qmando:
I've added a fix for this issue. See #593

326TimesBetter:
That definitely solved that strange issue. Now what happens when I run the script is I get this error:

[...]

PR #593, opened 2016-06-21
Fixed strings in command

Qmando:
Fixes an issue where if you don't have a "%" in the command, it will interpret a string as a list.
In alert(), we treat the command as a list, but it was only being converted from a string to a list IF there was a % formatter.

See #592


danielpops:
lgtm

y-trobinso:
lgtm

PR #603, opened 2016-06-29
Add slack channel config

SeanNguyen:
Added configuration option to customise Slack channel as this is very easy to achieve. Our team at Paypal really want this option to be enabled.

So that we can do something like this:
```yaml
alert:
- slack:
    slack_webhook_url: "https://hooks.slack.com/services/xxxxxx"
    slack_channel_override: "#some-channel"
    alert_text: "Hello World"
```


chernjie:
Woohoo! 👍

PR #610, opened 2016-07-01
Fixed run_enhancements_first bug

Qmando:
The ``continue`` command was in the context of the enhancements for loop. So instead, we catch the exception outside the loop and continue in the correct context.


y-trobinso:
lgtm. This change may desire tests.

danielpops:
ship it!

PR #618, opened 2016-07-07
Added option for silencing rules before enhancements run

Qmando:
~~Adds an option so that you can silence rules after a match even if the enhancements drop the match.~~

run_enhancements_first runs AFTER silence stash is created

### DIFF
--- a/docs/source/ruletypes.rst
+++ b/docs/source/ruletypes.rst
@@ -417,7 +417,8 @@ run_enhancements_first
 ^^^^^^^^^^^^^^^^^^^^^^
 
 ``run_enhancements_first``: If set to true, enhancements will be run as soon as a match is found. This means that they can be changed
-or dropped before affecting realert or being added to an aggregation. (Optional, boolean, default false)
+or dropped before affecting realert or being added to an aggregation. Silence stashes will still be created before the
+enhancement runs, meaning even if a ``DropMatchException`` is raised, the rule will still be silenced. (Optional, boolean, default false)
 
 query_key
 ^^^^^^^^^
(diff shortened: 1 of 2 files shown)


docs/source/ruletypes.rst
```diff
@@ -419,6 +419,12 @@ run_enhancements_first
 ``run_enhancements_first``: If set to true, enhancements will be run as soon as a match is found. This means that they can be changed
 or dropped before affecting realert or being added to an aggregation. (Optional, boolean, default false)
 
+silence_matches_first
+^^^^^^^^^^^^^^^^^^^^^
+
+``silence_matches_first``: If set to true, a silence stash will be created for matches before enhancements are run. This means that even
+if a ``DropMatchException`` is raised, the rule will be silenced.
+
```
danielpops: add "(Optional, boolean, default false)"

danielpops: I would also add more clarity on when you might want to use this flag.

docs/source/ruletypes.rst
```diff
@@ -419,12 +419,6 @@ run_enhancements_first
 ``run_enhancements_first``: If set to true, enhancements will be run as soon as a match is found. This means that they can be changed
 or dropped before affecting realert or being added to an aggregation. (Optional, boolean, default false)
```
danielpops: I would add some more color here to indicate the nuance of the behavior we discussed.

danielpops:
Other than a minor doc update to describe the nuance, lgtm.  ship-it!

danielpops:
sheep eat!

Issue #613, opened 2016-07-05
What


LazyerHong:
ERROR:root:Traceback (most recent call last):
  File "./elastalert.py", line 764, in run_all_rules
    num_matches = self.run_rule(rule, endtime, self.starttime)
  File "./elastalert.py", line 572, in run_rule
    if self.is_silenced(rule['name'] + key) or self.is_silenced(rule['name']):
  File "./elastalert.py", line 1264, in is_silenced
    self.handle_error("Error while querying for alert silence status: %s" % (e), {'rule': rule_name})
  File "/usr/lib/python2.6/site-packages/elasticsearch-2.3.0-py2.6.egg/elasticsearch/exceptions.py", line 55, in __str__
    cause = ', %r' % self.info['error']['root_cause'][0]['reason']
TypeError: string indices must be integers

ERROR:root:Uncaught exception running rule Example keywords rule: string indices must be integers
INFO:elastalert:Rule Example keywords rule disabled
Traceback (most recent call last):
  File "./elastalert.py", line 1409, in <module>
    sys.exit(main(sys.argv[1:]))
  File "./elastalert.py", line 1406, in main
    client.start()
  File "./elastalert.py", line 727, in start
    self.run_all_rules()
  File "./elastalert.py", line 782, in run_all_rules
    logging.warning("Querying from %s to %s took longer than %s!" % (old_starttime, endtime, self.run_every))
UnboundLocalError: local variable 'old_starttime' referenced before assignment

Qmando:
The first error is in the elasticsearch-py library. I'm not sure what the underlying cause of that is. You are using version 2.3.0, so you should also be using Elasticsearch 2.X. If you are using 1.X, use the 1.X version of the library. Try something like ``pip install elasticsearch-py==1.9.0`` if that's the case. 

As for the second one, that is an actual bug. I'll fix this soon. You could work around this temporarily by replacing that line (elastalert.py:782) with ``pass``.

PR #619, opened 2016-07-07
Fixed bug where old_starttime was unset

Qmando:
old_starttime is set only if there are no exceptions. We generally don't need to warn about a slow run if there was exceptions anyway. This moves the warning block into the try..else block.

Fixes part of #613

### DIFF
--- a/elastalert/elastalert.py
+++ b/elastalert/elastalert.py
@@ -772,13 +772,13 @@ class ElastAlerter():
                                                             self.num_hits, num_matches, self.alerts_sent))
                 self.alerts_sent = 0
 
-            self.remove_old_events(rule)
+                if next_run < datetime.datetime.utcnow():
+                    # We were processing for longer than our refresh interval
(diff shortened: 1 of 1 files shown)


elastalert/elastalert.py
```diff
@@ -772,13 +772,13 @@ def run_all_rules(self):
                                                             self.num_hits, num_matches, self.alerts_sent))
                 self.alerts_sent = 0
 
-            self.remove_old_events(rule)
+                if next_run < datetime.datetime.utcnow():
+                    # We were processing for longer than our refresh interval
+                    # This can happen if --start was specified with a large time period
+                    # or if we are running too slow to process events in real time.
+                    logging.warning("Querying from %s to %s took longer than %s!" % (old_starttime, endtime, self.run_every))
```
danielpops: Can't the value still be undefined if an exception happens?

danielpops:
lgtm. ship it

PR #654, opened 2016-07-27
Support timestamp field with dot

RubieV:
Uses lookup_es_key to report the timestamp from a Match, instead of failing lookup in nested dictionary.


Qmando:
Are these the only two places that call match[timestamp_field]? I'm actually a little surprised! Anyway, great stuff, thanks.

Issue #401, opened 2016-02-17
Rule type that alerts on average or sum of a field?


anhlqn:
Hi,

I wonder if ElastAlert has a rule type that alerts on average or sum of a field. With IIS Logs in ES for example:
Alert if average time-taken within 5 minutes is higher than 1000 ms.
Alert if sum of sc-bytes within 5 minutes is larger than 1,000,000,000 Bytes.

I'm looking for something similar to average and sum aggregation on Kibana, and it would be great if ElastAlert has such rules.

Thanks,
Anh

errordaiwa:
I have looked through ElastAlert config sheet and source code. It seems ElastAlert doesn't support aggregation query from ElasticSearch.

mr2china:
I also have this requirement ,if have general solution for sum,count,average aggr on multi field of each document  ....it will be  powerful

errordaiwa:
@mr2china I made a [patch](https://github.com/errordaiwa/elastalert/tree/feature/elasticsearch_aggregations_query_support) to support these features. The implement seems a little tricky since I'm not familiar with python.

mr2china:
@errordaiwa ,I only implement by using  rule extending way,like sum (divisor)/sum (dividend),which in () is multi-fields to do +,-,*,/   in each document.

mr2china:
adding, we know es support script like groovy ,it's powerful ,so we have to implement by   
python to simulate groovy or other script.

abiruth84:
Hi! 

I was actually wondering the same posted here. 
I'm trying to set up a couple of alerts based on the average of 2 fields (requests_queued and time_taken) for a certain period of time and I am not sure how to do it. 

@errordaiwa you mentioned a patch to support this but I am not sure yet how to implement it. 
Thanks,
Ruth

errordaiwa:
@abiruth84 add configuration in yaml like this

```
aggs:
  aggs_key: $key_name
  aggs_form:
    $key_name:
      avg:
        field: requests_queued
```

abiruth84:
@errordaiwa sorry to ask again. 
From the patch you linked, I just need to add these files that you modified to my current EA version? 
- elastalert/elastalert.py
- elastalert/ruletypes.py
- elastalert/util.py 

and then in my rule.yaml I add the "aggs" part? Is this "aggs" the rule "type"? do you maybe have some example documented on this change? 
sorry but my knowledge with python and the insights of elastalert code is a bit limited at this point. 
[...]

PR #657, opened 2016-07-28
Issue #401: support aggregation query from elasticsearch

errordaiwa:
See #401 . Support aggregation query like sum, average and etc.


pataquets:
@errordaiwa : #846 is in. Which are the differences/improvements that this PR provides? Is this PR still needed/can be closed?

PR #658, opened 2016-07-28
Allow nested filter with change in schema

RubieV:
Add support for nested filter [ES Nested Filter DSL](https://www.elastic.co/guide/en/elasticsearch/reference/2.0/query-dsl-nested-query.html) 

### Example:
```
filter:
  - nested:
      path: nested
      query:
        exists:
          field: nested
```

Possible side effect might be dropped support for:
```
filter:
  term: a
```
This will not work as the filter object is now a dict. Might be fixed with more advanced schema.

### DIFF
--- a/elastalert/schema.yaml
+++ b/elastalert/schema.yaml
@@ -145,7 +145,7 @@ properties:
   filter :
     type: [array, object]
     items: *filter
-    additionalProperties: false
+    additionalProperties: true
     properties:
         download_dashboard: {type: string}
 


Qmando:
Isn't you example valid under current schema?

In
```
filter:
 - nested:
```
filter is a list, so additionalproperties shouldn't affect it.

Where as
```
filter:
  term: a
```
is NOT currently valid.

I've never used nested queries, but I feel like it should work as it currently is. Take https://www.elastic.co/guide/en/elasticsearch/guide/current/nested-query.html for example. You could create this by using 

```
filter:
  - match:
       title: eggs
  - nested:
       path: "comments"
       query:
         bool:
           must: [ ... ]
```

RubieV:
You are right, my tests were off. I might come back and fix that line, sometime.

PR #671, opened 2016-08-10
update configparser version

ltagliamonte:
Had a problem installing on aws ami with python 2.7
This PR allows to use newer configparser versions.


Jonnymcc:
+1

PR #707, opened 2016-09-12
Don't exit if rule directory is empty on startup

RubieV:
Currently, if starting Elastalert,  load_rules exits when the rule directory is currently empty. This behavior is unwanted, as Elastalert watches for new files in the rules directory, and will pick them up when available. Now, running Elastalert as daemon is unpredictable in case the rule directory is empty. 

The current fix removes the exit behavior on startup when the rule directory is empty.


bean5:
I agree with this.

PR #715, opened 2016-09-20
Stomp protocol added in order to notify brokers such as ActiveMQ

snuids:
A new type of alerter added in order to send alerts to brokers such as ActiveMQ or RabbitMQ.

Will probably add the AMQP protocols which is widely used with classic middleware.

Example of alert configuration:
# (Required)
# The alert is use when a match is found

alert:
- "stomp"
# (required, stomp specific)

stomp_hostname: 127.0.0.1
stomp_hostport: 61613
stomp_login: admin
stomp_password: admin
stomp_destination: /queue/ALERT3


elastalert/util.py
```diff
@@ -8,7 +8,7 @@
 from elasticsearch import RequestsHttpConnection
 from elasticsearch.client import Elasticsearch
 
-logging.basicConfig()
+logging.basicConfig(level=logging.INFO, filename='./alerter.log')
```
Qmando: I think I would rather see file logging as a setting, not the default.

snuids:
How do you want the documentation. Do you want me to prepare a text file compatible with the current doc ? (https://github.com/Yelp/elastalert/edit/master/docs/source/ruletypes.rst)

elastalert/util.py
snuids: I will remove that

snuids:
Can I do something to close the pull ?

PR #718, opened 2016-09-20
Dpopes sec 5621 keyed time aggregations

danielpops:
Sending out in a kind of WIP state:

-  Existing tests pass (with minor changes or additions)
-  Did some preliminary testing and am seeing the correct behavior w.r.t. alerts being sent out in groups based on the key
-  I would like to fix the "first record" hack. @Qmando looking at you for suggestions. I *think* I can just tag even the first encountered record with it's own ID as aggregate_id (and optionally aggregate_key) and change downstream ES queries accordingly.
-  I need to add more unittest coverage for the new functionality.
-  This won't work as-is for a composite query_key field, nor will it work for a non-primitive value of the query_key dimension (for the former, I need to look closer and think about what the solution looks like. for the latter, this can be solved by hashing the contents of the field). It doesn't look a composite query_key field is supported correctly today for realert.

### DIFF
--- a/elastalert/elastalert.py
+++ b/elastalert/elastalert.py
@@ -1046,8 +1046,8 @@ class ElastAlerter():
 
             # Send the alert unless it's a future alert
             if ts_now() > ts_to_dt(alert_time):
-                aggregation_keys = self.get_aggregated_match_keys(_id)
-                for key in aggregation_keys:
+                aggregate_keys = self.get_aggregated_match_keys(_id)
+                for key in aggregate_keys:
                     aggregated_matches = self.get_aggregated_matches(_id, key)
                     if aggregated_matches:
                         matches = [match_body] + [agg_match['match_body'] for agg_match in aggregated_matches]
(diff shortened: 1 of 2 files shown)


elastalert/elastalert.py
```diff
@@ -1107,6 +1170,11 @@ def find_pending_aggregate_alert(self, rule):
 
     def add_aggregated_alert(self, match, rule):
```
Qmando: What if get_aggregated_matches and send_pending_alerts were left (almost) verbatim. And the following changes were added into add_aggregated_alert:

rule['current_aggregate_id'] becomes rule['current_aggregate_id'].get(query_key_value)
rule['aggregate_alert_time'] becomes rule['aggregate_alert_time'].get(query_key_value)

and the relevant code is updated in other places to initialize this object to a dictionary.

danielpops:
disregard this PR. Instead, look at https://github.com/Yelp/elastalert/pull/721

PR #721, opened 2016-09-21
keyed aggregations (simpler approach)

danielpops:
Thanks @Qmando for the review comments yesterday.  You were right, it is possible and much better to simplify as per your original suggestion.

### DIFF
--- a/elastalert/elastalert.py
+++ b/elastalert/elastalert.py
@@ -457,6 +457,20 @@ class ElastAlerter():
             return rule.get('buffer_time', self.buffer_time)
         return self.run_every
 
+    def get_query_key_value(self, rule, match):
+        # concatenate query_key (or none) with rule_name to form the key used for silence_cache, grouped aggregates, etc.
+        if 'query_key' in rule:
+            try:
+                key = unicode(lookup_es_key(match, rule['query_key']))
+            except KeyError:
+                # Some matches may not have a query key
+                # Use a special token for these to not clobber all alerts
+                key = '_missing'
+        else:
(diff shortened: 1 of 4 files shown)


danielpops:
I realized that I should also add some doc updates.  Another iteration coming soon.

elastalert/elastalert.py
```diff
@@ -509,25 +527,18 @@ def run_rule(self, rule, endtime, starttime=None):
...
-                except KeyError:
-                    # Some matches may not have a query key
-                    # Use a special token for these to not clobber all alerts
-                    key = '._missing'
-            else:
-                key = ''
-
-            if self.is_silenced(rule['name'] + key) or self.is_silenced(rule['name']):
-                elastalert_logger.info('Ignoring match for silenced rule %s%s' % (rule['name'], key))
+            if self.is_silenced(silence_cache_key):
```
Qmando: I would like to keep the additional check for is_silenced(rule['name']). This allows yo to silence an entire rule, instead of only being able to silence one value of query_key.

danielpops: But the silence_cache_key will either be the rule name alone OR it will be the rule name + querykey (if query_key is set).  In what scenario will elastalert try to silence the whole rule, but not specific keys, if query_key is set?

Qmando: If you manually run --silence

danielpops: Addressed this by identifying entries silenced by --silence differently than entries silenced through the realert parameter.

PR #722, opened 2016-09-21
Typo fixes and normalization

bean5:
Make some words more consistent and fixed a typo.

Just let me know if any of these aren't in-line with what you would like. I am open to changing it.


Qmando:
Elastic doesn't capitalize the S in Elasticsearch. See https://www.elastic.co/

bean5:
At least the codebase is consistent now...I'll take a look.

bean5:
How about now, @Qmando ?

PR #723, opened 2016-09-22
Avoid confusion in systems that have write-back configuration

bean5:
I was confused when I first saw this message because I have a custom writeback index--and I thought that something might be misconfigured. Turns out the verbiage just needed clarification.

### DIFF
--- a/elastalert/test_rule.py
+++ b/elastalert/test_rule.py
@@ -234,7 +234,7 @@ class MockElastAlerter(object):
             client.run_rule(rule, endtime, starttime)
 
             if mock_writeback.call_count:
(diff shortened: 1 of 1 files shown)


bean5:
Hmm. Failed the first time I pushed this. Second time it worked.

Compare
https://travis-ci.org/Yelp/elastalert/builds/161789030
https://travis-ci.org/Yelp/elastalert/builds/161907270

Looks like a test is finnicky.

Qmando:
@bean5 I just fixed that flakey test. It relied on dictionary ordering.

bean5:
Cool.

PR #735, opened 2016-09-27
summary table for aggregations

danielpops:
With aggregations, there can be many documents in an alert email/jira/etc.
Adding support for a summary table that can show a summary of relevant fields that the alert author may be most interested in.


Qmando:
Shipit

PR #740, opened 2016-09-29
Factor in query delay for remove_old_events

### LINKED ISSUE #739: Duplicate alerts with query_delay
When using a query_delay without use_count_query or use_terms_query, there will sometimes be duplicate events. This is because in remove_old_events, ts_now() is used to calculate which documents are out of the current query range, and will forget events as soon as they are older than present time minus buffer time. However, with query_delay, these forgotten events can still be queried and cause duplicate alerts.

Qmando:
Fixes #739 

Don't forget _ids until they are buffer_time PLUS query delay old.


danielpops:
lgtm.

Ship it!

PR #746, opened 2016-09-30
Added sort to get_aggregated_matches

Qmando:
Grab the aggregate matches in ascending time order so that they will already be in order when the alerter receives them.

### DIFF
--- a/elastalert/elastalert.py
+++ b/elastalert/elastalert.py
@@ -1105,7 +1105,7 @@ class ElastAlerter():
         """ Removes and returns all matches from writeback_es that have aggregate_id == _id """
 
         # XXX if there are more than self.max_aggregation matches, you have big alerts and we will leave entries in ES.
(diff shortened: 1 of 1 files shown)


elastalert/elastalert.py
```diff
@@ -1105,7 +1105,7 @@ def get_aggregated_matches(self, _id):
         """ Removes and returns all matches from writeback_es that have aggregate_id == _id """
 
         # XXX if there are more than self.max_aggregation matches, you have big alerts and we will leave entries in ES.
-        query = {'query': {'query_string': {'query': 'aggregate_id:%s' % (_id)}}}
+        query = {'query': {'query_string': {'query': 'aggregate_id:%s' % (_id)}}, "sort": {"@timestamp": "asc"}}
```
danielpops: The other dictionary fields have single quotes. You should have these match for consistency sake

PR #756, opened 2016-10-04
Allow for setting hipchat from attribute

bean5:
I am new to python and to ElastAlert, so I didn't so any bounds checking. This code assumes that users won't set a string that is longer than 64 characters. See https://www.hipchat.com/docs/apiv2/method/send_room_notification. Empty string is allowed according to that page.

@Qmando: Is there an easy way to run tests on this project, perhaps in a docker container. I prefer not to install python or any other language on my system (docker is good at doing that on a per-project level). 

I saw the tox target in the Makefile, but my computer certainly doesn't have that installed either. Any chance you could outline how to get a test environment set up? I could probably whip up a docker target for running tests.

PR #758, opened 2016-10-05
Add framework for running tests in a docker container

bean5:
Requires docker-compose.

### DIFF
--- a/Dockerfile-test
+++ b/Dockerfile-test
@@ -1,0 +1,11 @@
+FROM ubuntu:latest
+
+RUN apt-get update && apt-get upgrade -y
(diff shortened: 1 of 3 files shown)


bean5:
I get a blist error like the one you get at https://github.com/Yelp/elastalert/pull/756#issuecomment-251522025. Do you know what dependency needs to be installed, @Qmando?

bean5:
I didn't add a docker-compose target for starting this as a service, but it wouldn't be a bad idea. It would remove my dependence on https://hub.docker.com/r/ivankrizsan/elastalert/ or https://github.com/krizsan/alerting-with-elk-and-elastalert.

PR #799, opened 2016-11-07
Use threading module to run queries in parallel

### LINKED ISSUE #17: Using threading for to avoid blocking on ES queries
Rewrite the rules loop to use Python threading facilities, allowing rules to be run in parallel in multicore environments. Inspect the performance implications of this and correct for any lock contention or global state issues that may arise.

lucasrodcosta:
Closes #17.  
Uses `threading` module to run Elasticsearch queries in parallel.

### A little benchmark
The table below shows a little benchmark that I ran in my machine (a [MacBook Pro 2011](https://support.apple.com/kb/SP619)).  
This benchmark is composed by 8 rules used in the production environment of the company that I work.

Each row of the table refers to a different configuration, and the first row corresponds to the original elastalert implementation (without threads).

For each configuration I ran the elastalert 10 times and measure the time spent (in seconds) in each execution.  
Then I finally calculated the [mean](https://en.wikipedia.org/wiki/Arithmetic_mean), the [standard deviation](https://en.wikipedia.org/wiki/Standard_deviation) and the [variance](https://en.wikipedia.org/wiki/Variance) and put the results in the table.

|                 | mean (sec) | standard deviation (sec) | variance (sec) | difference (%) |
|:---------------:|:----------:|:------------------------:|:--------------:|:--------------:|
| without threads | 6.5513104  | 0.3732684                | 0.13932931     | N/A            |
| 2 threads       | 6.5522794  | 0.2512189                | 0.0631109      | +  0.0148%     |
| 4 threads       | 4.4218844  | 0.90058841               | 0.81105948     | - 32.5038%     |
| 8 threads       | 4.0367426  | 0.12913148               | 0.01667494     | - 38.3827%     |

As show above, the code rans 38% faster using 8 threads.

### DIFF
--- a/elastalert/elastalert.py
+++ b/elastalert/elastalert.py
@@ -9,6 +9,7 @@ import sys
 import time
 import traceback
 from email.mime.text import MIMEText
+from multiprocessing.pool import ThreadPool
 from smtplib import SMTP
 from smtplib import SMTPException
 from socket import error
@@ -102,6 +103,7 @@ class ElastAlerter():
         self.from_addr = self.conf.get('from_addr', 'ElastAlert')
         self.smtp_host = self.conf.get('smtp_host', 'localhost')
         self.max_aggregation = self.conf.get('max_aggregation', 10000)
+        self.num_workers = self.conf.get('num_workers', 1)
         self.alerts_sent = 0
         self.num_hits = 0
         self.current_es = None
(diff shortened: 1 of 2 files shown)


lucasrodcosta:
@Qmando, what do you think about this?
Makes sense?

Qmando:
I'm sorry about being super slow to review this. It does make a lot of sense. I promise I'll get it done this week.

elastalert/elastalert.py
```diff
@@ -719,6 +721,39 @@ def start(self):
...
+            num_matches = self.run_rule(rule, endtime, self.starttime)
+        except EAException as e:
+            self.handle_error("Error running rule %s: %s" % (rule['name'], e), {'rule': rule['name']})
+        except Exception as e:
+            self.handle_uncaught_exception(e, rule)
+        else:
+            old_starttime = pretty_ts(rule.get('original_starttime'), rule.get('use_local_time'))
+            elastalert_logger.info("Ran %s from %s to %s: %s query hits, %s matches,"
+                                   " %s alerts sent" % (rule['name'], old_starttime, pretty_ts(endtime, rule.get('use_local_time')),
+                                                        self.num_hits, num_matches, self.alerts_sent))
```
Qmando: self.num_hits needs to be mapped per rule. (It should have been in the first place!)

Qmando:
Awesome stuff! I've moved num_hits into rule dictionaries, ran some tests, and am moving the PR to https://github.com/Yelp/elastalert/pull/845

PR #821, opened 2016-11-22
Use all config.yaml defaults

Qmando:
Instead of picking a choosing defaults, just use everything

### DIFF
--- a/elastalert/config.py
+++ b/elastalert/config.py
@@ -66,6 +66,8 @@ alerts_order = {
     'email': 1
 }
 
+base_config = {}
+
 
 def get_module(module_name):
     """ Loads a module and returns a specific object.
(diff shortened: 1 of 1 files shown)


elastalert/config.py
```diff
@@ -133,7 +135,9 @@ def load_options(rule, conf, args=None):
     except (KeyError, TypeError) as e:
         raise EAException('Invalid time format used: %s' % (e))
 
-    # Set defaults
+    # Set defaults, copy defaults from config.yaml
+    for key, val in base_config.items():
```
danielpops: Shouldn't you do this for after the other rule.setdefault lines below? If somebody were to set their timestamp globally in config.yaml, you would overwrite it on line 147.  Is that a supported scenario to put arbitrary rule parameters like that in config.yaml?

Qmando: Setdefault means it won't change the value if it already exists. ``rule.setdefault`` can never overwrite anything.

PR #841, opened 2016-12-08
Changes to get count and terms queries working with es_support branch.

mircopolo:
If you are interested some changes to get count and terms queries on the es_support branch. Both types of query (use_count and use_terms) are working fine my es 5.0 but I don't have an es 2x to hand for regression testing.


Qmando:
Awesome!! I'm pretty sure this shouldn't change behavior at all when self.five is False, so it LGTM!

Issue #727, opened 2016-09-23
frequency rule that uses sum total of a numeric field value instead of event counts?


PMDubuc:
I'm trying to implement a rule similar to frequency that adds up the values of a particular number field in each event that matches a query and alerts on the value of that sum within a timeframe instead of the number of events.  I'm trying to read the frequency rule source and the documentation example on adding a new rule type but my python is a bit rusty and I'm not sure where to start.

Can anyone give me some pointers or an example?
Thanks.

PMDubuc:
I was able to work up a modified version of the frequency rule source to do what I need.

MrGlobi:
Hi,
Im trying the same, can you share your configuration?

ruget:
+1, I'm looking for the same functionality

Issue #282, opened 2015-10-21
Alert based on % of a query related to total number of events


ricardojdsilva87:
Hello,

Currently I'm trying to generate an alert to generate an email when the access success rate on a website is below xx%. The data I'm indexing is based on apache logs, and I'm interested on the field that has the HTTP code (200,404, 500, etc).

I was able to generate a dashboard that shows the access success rate, based on the following queries:
response: [ 0 TO 499 ]
response: [ 500 TO \* ]

The graph shows basically which percentage of calls have HTTP code <=500 compared to the rest of the events. It looks something like this:
![image](https://cloud.githubusercontent.com/assets/12953702/10644834/9264e664-7822-11e5-9687-59c3afde5570.png)

For this I had to select the percentage option on the view:
![image](https://cloud.githubusercontent.com/assets/12953702/10644862/ae4fda64-7822-11e5-92fb-fa283bf9a616.png)

I'm using the filter type frequency, and on the query currently I have this
- query:
  query_string:
    query: "response:[500 TO *]"

What this does is to count the number of 500+ requests, also the Cardinality option would work, because it would aggregate different HTTP codes (200,404,500,etc) and not in a range [0 to 499] and [500 to *].

My question is, if there is any option to do this using the query from the kibana dashboard and this way have an alert sending mails when in the last 5 minutes the threshold passed below 95% for example.

Thanks for the help

Qmando:
Unfortunately there is not currently any % based rule types. You can either use a fixed threshold, with ``frequency``, or alert if the number of 5XX responses increases by X%, using ``spike``. The cardinality rule could also do something like "alert if more than 10 users receive 5XX responses within an hour".

I would really like to add this feature for ratios of results of different queries.

ricardojdsilva87:
Hello, thanks for the quick reply,

I think I can work with the spike option, it's just a matter of checking the "normal" number of 5XX events to define the threshold.

I think that the % option would be a great improvement to the tool. Keep the great work.
Should I close the issue? Or do you want to keep it to be an enhancement? I changed the title also to something more related to the question.

Thanks again

stefanrehm:
@qmando :  Do you know if there was any progress on this feature? It is something that we would love to use as well.. ;)

codinguari:
Any progress on this feature. I would love to use something like this

redhelling21:
I actually had the same problem, and i did something like this, following the custom rules part of the doc : 

```
import dateutil.parser
[...]

PR #846, opened 2016-12-15
Experimental single metric aggregation rule for Elastalert.

mircopolo:
**Updated 20/02/2017**
Hi just wanted to start some discussion on a single value metric aggregation that I have prototyped. It is very much a work in progress that will need plenty of cleanup + tests, but I wanted to get some early feedback before putting in more work, in case a) there is something else in development already or b) you are not keen on the approach I have taken.

I have seen a similar PR that extends Frequency and Count rules and this is not dissimilar, but I don't think the non overlapping nature of the current document counting rules is the best fit for metric aggregations (One might want to check CPU usage over the past five minutes every minute,  etc) so I would like to provide an option for an overlapping buffer times (NB: I haven't fully implemented this particular requirement just yet as so far I have been using test rule). 

The basic approach is basically as follows:

- required: **metric_agg_key** - a field to perform the (single-value) metric aggregation on
- required:  **metric_agg_type** - one of (min, max, sum,avg,value_count, cardinality). I'd love to add others such as stats, extended stats and percentiles but haven't yet found of a nice way to express threshold #configuration for these.
- required (one of): **min_threshold**, **max_threshold**
An alert will triggered if min or max threshold is breached.

By default it will run the query over **buffer_time**. Setting **use_run_every_query_size** to true will default to using run_every instead as the default segment size.

It will also respect the current **query_key** + **doc_type** functionality, and where this is    ###provided it will add the single value metric as a sub aggregation of each term bucket. 

Where it gets a little more complicated ....

I have added a further time delta parameter **bucket_interval**, where this is included the metric aggregation, or term aggregation (if use_query_key) will become a sub aggregation of a date histogram broken down into **bucket_interval** sized time interval over **buffer_time** as a whole. if **bucket_interval** is specified, then **buffer_size** must be evenly divisible by **bucket_interval** (or if **use_run_every_query_size** has bee set to true then **run_every** must be evenly divisible by **bucket_interval**) .

An additional parameter **sync_bucket_interval**,  may be applied if **bucket_interval** has been set . As background for a given interval in a date histogram elasticsearch calculates the bucket keys (times) from the unix epoch. As a result you can end up with of partial buckets when the start and end times don't align with elastalert start and end times. 

So the default behavior is to push the bucket keys forward to align with the query start time (this is achieved by calculating the appropriate offset parameter for the data histogram to align exactly with the start date). When sync_bucket_interval is true, the start and end time will be adjusted backwards to align with the bucket date keys from elasticsearch. When sync_bucket_interval is false, the bucket date keys times will aligned with the start and end times of the elastalert query by applying an offset. 

The **allow_buffer_time_overlap** applies if using **buffer_time** as the segment size and  **buffer_time** is greater than run_every interval. When this is set to true, the start time of subsequent queries will go back beyond the end time of the previous query. This supports more advanced use cases such as checking cpu% usage over the past 5 minutes every minute. 

Although I just adding a single value metric rule, I feel this approach could easily be extended to more complex rules such as min occurrences, spikes etc. The outer date_histogram, and terms aggregations are fixed (but optional components) where as the 'payload' sub-aggregation, which my case is a single-value metrics could potentially be any sort of aggregation result that subclass rules know how to process.

Cavets:
- Only used ES5 for testing so far (will need to test on other versions as this progresses)
- Only ran these using used elastalert-test-rule so, so there are probably a good few of restart/time scenarios that I have not yet covered in the full program. 

I probably won't get to come back to this until January but if I could get some feedback on the general direction that I am taking. I wouldn't have much python experience either, or feel I have a comprehensive understanding of elastalert as of yet so feedback on coding standards or quality are most welcome.

#401 
#657


cbarbaza:
Hi !

please keep going working on it, it's an interested feature that i really need :+1:

4admin2root:
just I wanted

cbarbaza:
Hi, 

is the pull request still waiting ?
This is a realy needed features !!!!! 
:100:

PMDubuc:
#727

mircopolo:
Pretty much feature complete in terms of readiness for code review.  
Major to do items are writing new tests and updating the documentation which I will hope to get back to shortly.

PMDubuc:
Thanks for your work on this!  I'm wondering if this will support multiple rule definitions that differ only in the time period and thresholds.  For example, can I have a rule that alerts on a metric if it exceeds a given threshold in 24 hours and another rule for the same metric if it exceeds a different threshold in 1 hour, etc.  One way to do this would be to use the name of the rule as part of the key for the aggregate data so the data for Rule.24hr would not conflict with that for Rule.1hr.

mircopolo:
@PMDubuc Not 100% sure I understand the use case fully, but I am guessing what you are asking for is the ability to set buffer_time to 24hrs, and bucket_interval to 1hr and have different thresholds for both the overall window and the sub intervals? As it stands you would need a completely separate rule to do this currently.

Changing the query wouldn't be much of a problem, it would just mean another aggregate over buffer_time at the same level as the date_histogram aggregation (like below), but for silences elastalert only uses the Rule name and query_key currently which is might be a blocker as the alerts for each would clash. 

It might also make the config a bit more complex, but it is a potentially powerful use case I hadn't considered that I will certainly have a think about. 

`
"aggs" : {

		"bucket_aggs" : {
			"terms" : {
				"field" : "beat.hostname",
				"size" : 50

			},
			"aggs" : {
				"interval_aggs" : {
					"date_histogram" : {
						"field" : "@timestamp",
						"interval" : "1h"
					},
					"aggs" : {
						"system_cpu_user_pct_avg" : {
							"avg" : {
								"field" : "system.cpu.user.pct"
							}
						}
					}
				},
				"total_system_cpu_user_pct_avg" : {
							"avg" : {
								"field" : "system.cpu.user.pct"
							}
				}
			}
		}
}`

**Results**

`
"aggregations" : {
  "bucket_aggs" : {
    "buckets" : [{
      "key" : "MY-HOSTNAME",
      "doc_count" : 6,
      "overall_system_cpu_user_pct_avg" : {
      "value" : 0.015333333333333332
      },
      "interval_aggs" : {
        "buckets" : [{
        "key_as_string" : "2017-02-20T16:37:00.000Z",
        "key" : 1487608620000,
        "doc_count" : 1,
        "system_cpu_user_pct_avg" : {
        "value" : 0.011
        }
      }]
      }
   }]
  }
}	  
`

mircopolo:
Closed by accident

leonhartX:
Really nice work!
I have add a support for `percentiles` and `percentile_ranks` based on your previous commits [here](https://github.com/leonhartX/elastalert/commit/8b0a4d15a1287c75f52f4bafd6c9630fbc7919a4).
It's a simple approach to check the percentile or rank with one value/threshold, since I need to check something like 'the response with status code 500 is over 1% in last 5 mins'
Would you consider adding this feature to the PR?

mircopolo:
@leonhartX Sure it looks good to me, I had intended to come back to the percentiles later, I guess the the easiest thing would be to open a PR to the the aggregation_metrics branch of my fork and I'll merge it in there? (I'm still newish to Github workfow so if there is a better way let me know) I have made a few commits since then but there shouldn't be any major conflicts from what I can see.

Qmando:
I'm gonna try to merge this into master today. Doing some testing now and it looks good.

mircopolo:
Whatever you think best yourself, but might a separate branch be better for a couple of weeks at least? Its just that I have done very little testing against anything other than ES 5.0. From what I have gleaned from the docs most simple single value metrics (max, min, sum, avg etc) ought to be compatible all the way back to 1.x

Either way I won't get the test completed just yet, but I'll try and get around to adding in some legible documentation at the very least in the next few days.

mircopolo:
@leonhartX: If this is merged to master guess you can just go straight there with any PR

Qmando:
I'm testing this against an Elasticsearch 1.X cluster and it works fine.

PMDubuc:
@mircopolo I think my scenario is simple.  If I want to use this rule type to alert on the sum of a field value during a 24 hour time period.  Can I also add another rule that will alert if the sum exceeds a different threshold over a different time period.  Will the sums be kept separate for each rule?  For example, if I want to keep track of use counts for a particular metric type for each userid, can I have a rule which sets an alerting threshold for a query key [metric, userid] for a 24 hour time period and another rule with the same query key but using a different time period and threshold?  My concern is that the use counts in the events will be added twice since both rules use the same query key.  I think this is a problem with the Frequency rule type, which counts the number of events instead of a field value in the events.  Thanks.

mircopolo:
@PMDubuc  Ah ok I understand now, while I not familiar with a query_key issue in other rule types, there is no reason why your scenario should work just fine. 

As currently implemented, unlike most of the 'document counting' type rules, the Metric aggregation rule has no need not need to save any intermediate state between query executions so there is no reason that two separate rules definitions should interfere with each other in any scenario.

mircopolo:
One last "Percentage match" rule type. It constructs a filters aggregation with a just two buckets, a match bucket and the _other_ bucket. It will calculate the number of documents in the match bucket vs the total number of document in match bucket + other bucket as a percentage, and alert if the percentage is below/above the min\max thresholds.

By my reckoning it should meet the requirement for 
#282
#911

mircopolo:
Any suggestion I will need to find a better place in the docs for use_run_every_query_size, allow_buffer_time_overlap, bucket_interval and sync_bucket_interval` setting as there will be not specific to the Metric Aggegartion rule.

Issue #553, opened 2016-05-31
Feature Request: Alerta Support


hrzbrg:
Hi,

it would be awesome to see Alerta as a build in alert output. It is an easy to use opensource alert dashboard with a simple API: http://alerta.readthedocs.io/en/latest/api/reference.html

Currently we made a workaround an let Elastalert trigger the Alerta CLI.

Regards
hrzbrg

minhdanh:
:+1:

jeff-cook:
Is anyone using this alert type? I see it has was added 2 years ago. Has anyone been successful with it?

Thanks

jurgenhaas:
Yes, I'm using it successfully and the alerta specific components in each role look like this:

```
alert:
  - alerta
alerta_api_url: "https://alerta.example.com/api/alert"
alerta_api_key: "THEKEYFROMALERTA"
alerta_environment: "name"
alerta_use_qk_as_resource: true
alerta_text: "Text of alert"
```

This works really well, the only issue I'm having is #2310 where I would like to set severity dynamically so that depending on the ES records I could open or close an alerta item.

moix:
I just integrated elastalert with alerta and first tests are working so far, so seems fine, thanks

abhishekjiitr:
Yes, the Alerta integration works as mentioned in the documentation, have been using it for several months now.

PR #848, opened 2016-12-22
Support alerta alerts

mircopolo:
New Alert for posting alerts to the /alert endpoint of the Alerta Server API. Supports both unauthenticated and authenticated scenarios via an optional Authorization header.

See also
#553


mircopolo:
Done under #1633

PR #853, opened 2016-12-29
Add support to replace dots in field names before writing to Elasticsearch

lperry:
When ElastAlert is used with Elasticsearch 2.x [1], elastalert writes to the writeback index fail when match_body contains fields with dots in their name (e.g., a frequency rule with a query_key containing dots).

Example Error

ERROR:root:Error writing alert info to Elasticsearch: TransportError(400, {u'root_cause': [{u'reason': u"Field name [d.ip.raw] cannot contain '.'", u'type': u'mapper_parsing_exception'}], u'type': u'mapper_parsing_exception', u'reason': u"Field name [d.ip.raw] cannot contain '.'"})

This PR adds support to replace any dots in field names with an underscore before writing to Elasticsearch. The default is off.

[1] https://www.elastic.co/guide/en/elasticsearch/reference/2.4/dots-in-names.html

### DIFF
--- a/docs/source/elastalert.rst
+++ b/docs/source/elastalert.rst
@@ -170,6 +170,9 @@ unless overwritten in the rule config. The default is "localhost".
 
 ``boto_profile``: Boto profile to use when signing requests to Amazon Elasticsearch Service, if you don't want to use the instance role keys.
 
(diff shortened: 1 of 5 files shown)


Qmando:
Ah, thank you so much! I'll try to review this in a somewhat timely manner.

PR #875, opened 2017-01-19
Be a little forgiving in filtering file extensions

etruong42:
.yml is a rather common file extension for YAML files. I ask that increase our tolerance of different ways of thinking and different, common file extensions.


Qmando:
Merged as https://github.com/Yelp/elastalert/commit/d6cd1e92857b336a7f016cb1d3fc746a897d2639

PR #876, opened 2017-01-21
add --config option to elastalert-create-index cli

Khezen:
I am storing tthe global config under `/etc/elastalert/elastalert.yml` and I need to run `elastalert-create-index` in a none interactive mode which is kind of hard to achieve because of `getpass.getpass` in the following block:
```
username = raw_input('Enter optional basic-auth username (or leave blank): ')
password = getpass.getpass('Enter optional basic-auth password (or leave blank): ')
```
Adding a `--config path/to/config` option is the most elegant solution I came up with.

### DIFF
--- a/elastalert/create_index.py
+++ b/elastalert/create_index.py
@@ -31,15 +31,13 @@ def main():
     parser.add_argument('--boto-profile', default=None, help='Boto profile to use for signing requests')
     parser.add_argument('--aws-region', default=None, help='AWS Region to use for signing requests')
     parser.add_argument('--timeout', default=60, help='Elasticsearch request timeout')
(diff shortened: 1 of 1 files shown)


elastalert/create_index.py
```diff
@@ -31,12 +31,13 @@ def main():
     parser.add_argument('--boto-profile', default=None, help='Boto profile to use for signing requests')
     parser.add_argument('--aws-region', default=None, help='AWS Region to use for signing requests')
     parser.add_argument('--timeout', default=60, help='Elasticsearch request timeout')
+    parser.add_argument('--config', default='config.yml', help='Global config file (default: config.yaml)')
```
Qmando: Can you change the default to config.yaml?

Khezen: of course

PR #656, opened 2016-07-27
Added a REST API to List, Create, and Test rules

CCoffie:
Below are some changes I made to ElastAlert that add a REST API. I added this because I've been working on a kibana plugin that I'll hopefully be releasing soon that will allow for quick and easy management of ElastAlert Rules from directly within kibana. Figured I'd push these now though because it might help other out there that want to do something similar.

PR #898, opened 2017-02-16
Add Kibana Plugin and Docker install

RubieV:
elastalertsamplerule

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,24 @@ Eg: ``--rule this_rule.yaml``
 
 ``--config`` allows you to specify the location of the configuration. By default, it is will look for config.yaml in the current directory.
 
+## Kibana Plugin and Docker install
+``./bin/kibana plugin -i elastalert -u https://git.bitsensor.io/front-end/elastalert-kibana-plugin/builds/5251/artifacts/file/build/elastalert-0.0.6.zip`` will install the Kibana plugin.
+
+Run and build the the Docker container with ElastAlert and RESTful frontend by
+```bash
+git clone https://git.bitsensor.io/back-end/elastalert.git && cd elastalert
+docker build . -t elastalert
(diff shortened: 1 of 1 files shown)


Qmando:
Pretty awesome!

kureus:
This looks really cool. Tho unfortunately the RESTful API part of the project only [officially supports](https://github.com/bitsensor/elastalert/blob/master/README.md#requirements) v0.0.96, which doesn't support elastic stack v5. A previous pull request had been opened to bring an API to the main repo (#656), which would avoid any version drift issues. @Qmando what are your thoughts on getting some form of http api inside the main repo?

RubieV:
Actually, we're running trunk so do support ES 5. Let me clarify:

Dockerfile:
RUN git clone https://github.com/Yelp/elastalert.git -b master --depth=1
/opt/elastalert

There is indeed risk on API change in master, and this is currently not
tested in CI. When this happens we will pin the last commit that we're able
to support, open an issue, and accept a pull to enable compatibility. Our
builds run on every commit, so that keeps backward compatibility with older
versions of ElastAlert.

Op ma 20 feb. 2017 om 14:37 schreef Dan Quirk <notifications@github.com>:

kureus:
@RubieV Thanks for the clarification. Did you consider looking to get the restful api inside the elastalert repo?

RubieV:
Yes, we chose an approach that would work with existing installations that would potentially have modifications or would have a specific version or would otherwise not be replaceable. As Node is already available due to Kibana, it is a lightweight approach.

Exposing the API from ElastAlert is better in terms of refactorability and tests are checked in with the code. I think that now we have provided the Kibana plugin, using the API that is defined makes it easy to move forward.

We use `master`, and will provide support for the `v0.1.8` API.

PR #909, opened 2017-02-20
Support aws credential providers

OrKoN:
Hi,

this PR allows configuring `aws_credentials_provider` to be either: `instance-metadata`(default) or `container`. If aws_credentials_provider equals `container`, the botocore.ContainerProvider is used to get aws_credentials (see http://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-iam-roles.html).

P.S. I could not figure out how to run tests locally so I hope Travis will notify me if something is wrong and I will fix it afterwards.

### DIFF
--- a/Dockerfile-test
+++ b/Dockerfile-test
@@ -1,8 +1,7 @@
 FROM ubuntu:latest
 
 RUN apt-get update && apt-get upgrade -y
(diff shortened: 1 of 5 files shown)


OrKoN:
1.

PR #820, opened 2016-11-22
Support es5

Qmando:
Initial es5 support.

create -> index everywhere.
Tests assume ES 2.
Queries use two formats, one for ES<5 and one for ES>=5.

Note:
-  test_rule.py not changed yet.
-  Old query string queries need to be changed.
-  Aggregations don't work yet.
-  Not very well tested, very much a work in progress

PR #910, opened 2017-02-21
Apply ES5 filter fix when using test rule

kureus:
When using elastalert-test-rule on elasticsearch v5 an error will be thrown that running your filter failed despite the test continuing to run successfully

image

It seems to keep backward syntax compatibility code was added to trim filters if using ES5 (https://github.com/Yelp/elastalert/pull/820/commits/bb09323ad893b22d17e0d7cb2aa50c52db18f80d) but this was only done inside the ElastAlerter class and not for MockElastAlerter. Not sure on preferred way to approach this but abstracted the code fix and used it in test_file() aswell fixing the error. Comments appreciated :-)

### DIFF
--- a/elastalert/elastalert.py
+++ b/elastalert/elastalert.py
@@ -657,15 +657,19 @@ class ElastAlerter():
                 continue
             new_rule[prop] = rule[prop]
 
-        # In ES5, filters starting with 'query' should have the top wrapper removed
         if self.is_five():
-            for es_filter in copy.copy(new_rule.get('filter', [])):
-                if es_filter.get('query'):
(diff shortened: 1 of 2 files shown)


Qmando:
Good stuff. Thanks a ton for fixing this!

Abhisar:
@kureus 
Could you please help me with a sample rule for ES 5.0.1 for frequency type.I have been trying but it seems the query isn't getting executed and all the records for frequency get alerted.

kureus:
@Abhisar You're best asking on the elastalert gitter for help. Or if you think there is a bug open an issue.

PR #920, opened 2017-03-01
Add SimpleAlerter for posting matches via http

muravitskiy:
This pull request adds new SimpleAlerter for posting unchanged matches and triggered rule name


elastalert/alerts.py
```diff
@@ -1178,3 +1178,35 @@ def alert(self, matches):
     def get_info(self):
         return {'type': 'ServiceNow',
                 'self.servicenow_rest_url': self.servicenow_rest_url}
+
+
+class SimpleAlerter(Alerter):
```
Qmando: Can we rename this SimplePostAlerter?

PR #921, opened 2017-03-01
allow reloading silence time from ES, when in-memory silence value is expired

ouchengeng:
Currently, elastalert.silence_cache is loaded only once when there's silence records in ES. After that, no matter what silence time is set, it will not be loaded into memory and thus it cannot take effect. But, we do alter or  lengthen silence time by calling 'elastalert --silence --rule' methods many times, especially when we have regular releases.  

In this pull request, I allow Elastalert to load silence time when in-memory silence value is expired.


Qmando:
Looks good to me, thanks!

PR #944, opened 2017-03-10
Added email_from_field feature

Qmando:
Add a feature to let you get the email recipient from a field in the alert event.

### DIFF
--- a/docs/source/ruletypes.rst
+++ b/docs/source/ruletypes.rst
@@ -1083,6 +1083,18 @@ This alert requires one additional option:
 
 Optional:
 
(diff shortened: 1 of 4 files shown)


Qmando:
For number 4, if you want to always include an email, you would use cc or bcc

nitin02:
How do I send to multiple recipients with this field? Can the value be a comma-separate list of emails?
(or)
Is there a way to add multiple fields in **email_from_field**. Each field has a single email specified?

I prefer the former as it is more dynamic.

pushkarevds:
Hello. How to send a message to default emails and email_from_field?

PR #947, opened 2017-03-13
Reference field "description" from alert type mail

felixrod78:
Hi all
I want to add a custom description in my alert, in order to appear in my json (elasticsearch), after hours of struggle , i can not find the way to do it.
Any idea how to do it?
Thanks


Qmando:
Did you mean to open an issue instead of a pull request?

What do you mean by "appear in my json (elasticsearch)"? You want this to be in the ``_type: elastalert`` document that's uploaded after the alert?

felixrod78:
Yes this is it

Qmando:
A quick way to solve this would be to add an extra field into the various alerters get_info methods. You could add ``"description": self.rule.get('alert_description')`` to the dictionary.

PR #950, opened 2017-03-13
Manage docs theme through pip

kureus:
Rather than rely on manually managing _static and overwriting the default theme, manage the choice of theme through a package so we aren't tethered to one point in time, presumably getting access to any theme updates/fixes. This approach does assume an unmodded theme.


kureus:
For reference the theme can be installed with 

```
pip install sphinx_rtd_theme
```

https://github.com/rtfd/sphinx_rtd_theme#installation

PR #957, opened 2017-03-17
Some minor enhancements

msmerc:
Make 'name' default to the filename.
Add the query json if a query fails.


elastalert/config.py
```diff
@@ -160,6 +164,7 @@ def load_options(rule, conf, args=None):
     # Set defaults, copy defaults from config.yaml
     for key, val in base_config.items():
         rule.setdefault(key, val)
+    rule.setdefault('name', os.path.splitext(filename)[1])
```
Qmando: This should be os.path.splitext(filename)[0]

>>> os.path.splitext('hello/world.yaml')
('hello/world', '.yaml')

msmerc: Doh!

PR #963, opened 2017-03-22
Improved test rule

msmerc:
- Made the config loading path much closer to that of elstalert
- If no rule is specified it will load and run all rules (though not very neatly)
- Added a --show-query option to print out the ES query
- Raises exceptions, so that error codes can be checked

### DIFF
--- a/elastalert/test_rule.py
+++ b/elastalert/test_rule.py
@@ -14,13 +14,10 @@ import sys
 
 import argparse
 import mock
-import simplejson
-import yaml
+import simplejson as json
 
-import elastalert.config
-from elastalert.config import load_modules
-from elastalert.config import load_options
-from elastalert.config import load_rule_yaml
+from elastalert.config import load_rules
+from staticconf.loader import yaml_loader
 from elastalert.elastalert import ElastAlerter
 from elastalert.util import elasticsearch_client
 from elastalert.util import lookup_es_key
@@ -47,9 +44,8 @@ class MockElastAlerter(object):
     def test_file(self, conf, args):
         """ Loads a rule config file, performs a query over the last day (args.days), lists available keys
         and prints the number of results. """
-        if args.schema_only:
-            return []
 
+        print("Testing rule: ", conf['name'])
         # Set up Elasticsearch client and query
         es_client = elasticsearch_client(conf)
 
@@ -58,7 +54,7 @@ class MockElastAlerter(object):
         except Exception as e:
             print("Error connecting to ElasticSearch:", file=sys.stderr)
             print(repr(e)[:2048], file=sys.stderr)
-            return None
+            raise e
 
         if is_five:
             ElastAlerter.modify_rule_for_ES5(conf)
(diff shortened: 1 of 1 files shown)


elastalert/test_rule.py
```diff
@@ -222,10 +217,9 @@ def get_id():
             endtime = ts_now()
             starttime = endtime - datetime.timedelta(days=args.days)
 
-        # Set run_every to cover the entire time range unless count query, terms query or agg query used
-        # This is to prevent query segmenting which unnecessarily slows down tests
-        if not rule.get('use_terms_query') and not rule.get('use_count_query') and not rule.get('aggregation_query_element'):
-            conf['run_every'] = endtime - starttime
+        # Set run_every to cover the entire time range, this is to prevent query segmenting which unnecessarily slows down tests
+        # (see elastalert.get_segment_size())
+        conf['run_every'] = endtime - starttime
```
Qmando: I don't think this should apply for use_count/terms_query. In those cases, it wont generate the same alerts as a normal run, because only a single query is made.

elastalert/test_rule.py
```diff
@@ -185,26 +183,23 @@ def mock_elastalert(self, elastalert):
...
+    def run_elastalert(self, conf, args):
         """ Creates an ElastAlert instance and run's over for a specific rule using either real or mock data. """
 
-        # Load and instantiate rule
-        load_modules(rule)
-        conf['rules'] = [rule]
-
-        # If using mock data, make sure it's sorted and find appropriate time range
-        timestamp_field = rule.get('timestamp_field', '@timestamp')
         if args.json:
```
Qmando: I'm not sure how this is supposed to work with JSON data AND multiple rules. things like timestamp_field can change per rule. I think if there are multiple rules it should not let you use json data, unless we can figure out a more elegant way to handle that.

elastalert/test_rule.py
msmerc: For the case where you have multiple rules it's not meanigful to say if you're using terms / count / agg (some rules might be, other rules might not). 

From a rough eyeball of the code, the `run_rule()` method that we use to test with only looks at `run_every` via `set_starttime()` and `get_segment_size()`. `set_starttime()` only uses `run_every` if the previous endtime is set, which it wouldn't be in this test (I think). `get_segment_size()` only applies to terms / count / agg queries anyway.

So basically I concluding that setting this parameter wouldn't do any harm.

elastalert/test_rule.py
msmerc: Good point - it wouldn't work for this case. I can try and fix it. Problem is that there aren't any tests in this part of the world so I'm wary of doing too much major surgery.

elastalert/test_rule.py
Qmando: I'm a little bit conflicted here, because seeing 
```
INFO:elastalert:Queried rule Example Rule from 2017-03-29 16:14 PDT to 2017-03-29 16:15 PDT: 0 hits
INFO:elastalert:Queried rule Example Rule from 2017-03-29 16:15 PDT to 2017-03-29 16:16 PDT: 0 hits
INFO:elastalert:Queried rule Example Rule from 2017-03-29 16:16 PDT to 2017-03-29 16:17 PDT: 0 hits
INFO:elastalert:Queried rule Example Rule from 2017-03-29 16:17 PDT to 2017-03-29 16:18 PDT: 0 hits
INFO:elastalert:Queried rule Example Rule from 2017-03-29 16:18 PDT to 2017-03-29 16:19 PDT: 0 hits
INFO:elastalert:Queried rule Example Rule from 2017-03-29 16:19 PDT to 2017-03-29 16:20 PDT: 0 hits
INFO:elastalert:Queried rule Example Rule from 2017-03-29 16:20 PDT to 2017-03-29 16:21 PDT: 0 hits
```
can be really slow and confusing to some people, it more mimics the real behavior. If someone has a rule with a small timeframe, this accuracy is crucial.

Seeing ``INFO:elastalert:Queried rule Example Rule from 2017-03-29 16:01 PDT to 2017-03-30 16:01 PDT: 0 hits`` is super quick, and useful to see the total number of matching documents, but won't trigger alerts in the same way.

The latter is almost equivalent to running it with --count-only. For non count/terms queries,  segment size doesn't matter, because each document is processed individually. That optimization was added to just make the output look nicer.

If we can't set run_every per rule, I'd rather see it not changed at all.

nsano-rururu:

PR #973, opened 2017-03-30
[Bug] Percentage Calculate is not True , and now recovery it

Zheaoli:
I find some bug in the Rule Percentage Rule.So I make a pr to recovery it


mircopolo:
What is the precise nature of the bug? You seem to be summing up the doc counts for each bucket_interval 
 sub bucket, which implies you want the total counts over buffer_time if I am correct?

However bucket_interval is an optional parameter so unless I am missing something you would just need to omit this from your rule definition to achieve the same behavior as the PR.

nsano-rururu:
Test are failing

nsano-rururu:

PR #976, opened 2017-03-31
Added support for rules with different versions

Qmando:
Add support for having different rules on different versions of elasticsearch. I used the following approach to make this change:

1. Change every self.is_five() to rule['five'] where rule is available and it's not forming a query for elastalert_status
2. Copy the code for version checking into modify_rule_for_ES5


elastalert/elastalert.py
```diff
@@ -775,13 +776,20 @@ def init_rule(self, new_rule, new=True):
...
 
     @staticmethod
     def modify_rule_for_ES5(new_rule):
+        # Get ES version per rule
+        rule_es = elasticsearch_client(new_rule)
+        if rule_es.info()['version']['number'].startswith('5'):
+            new_rule['five'] = True
+        else:
+            new_rule['five'] = False
+            return
```
danielpops: nit: unnecessary return

elastalert/elastalert.py
```diff
@@ -775,13 +776,20 @@ def init_rule(self, new_rule, new=True):
...
-        if self.is_five():
-            self.modify_rule_for_ES5(new_rule)
+        self.modify_rule_for_ES5(new_rule)
 
         return new_rule
 
     @staticmethod
     def modify_rule_for_ES5(new_rule):
+        # Get ES version per rule
+        rule_es = elasticsearch_client(new_rule)
```
danielpops: You could probably check for the presence of 'five' on the rule before creating the client and figuring out the version, but this might be an unnecessary micro-optimization.

elastalert/elastalert.py
Qmando: Except we now call modify_rule_for_es5 on every rule, and we don't want to pop the top level 'query' for non-five.

PR #1006, opened 2017-04-11
Use environment variables for configuration

micahhausler:
The following environment variables are supported in the main configuration, and override any values specified in the configuration:
- `ES_HOST`
- `ES_PORT`
- `ES_USERNAME`
- `ES_PASSWORD`
- `ES_USE_SSL`

I chose to have environment variables override configuration values, so that schema enforcement of the configuration doesn't change. This should probably be reworked at some point, so that schema enforcement can happen _after_ interpreting environment variables.


Qmando:
Schema enforcement seriously needs to be reworked! But this looks good for now.

Issue #711, opened 2016-09-16
Include blacklist and whitelist using files


pushkarevds:
Include blacklist and whitelist using files
rule
blacklist:
- include : /path/file.yaml

Will it be realized?

Qmando:
No

bean5:
An interesting idea. That would allow rules to share portions of what they match. I can see value in this. But formatting of the file would have be agreed upon.

SOCSIEM:
Need this feature too. ASAP. :)
Will be really useful to handle long list of malicious domains, URLs, IPs

SherifEldeeb:
I did a small modification to allow for this (still testing before submitting a pull request) ... it works for both blacklist and whitelist.
This will do two things:
- Add a new option to include entries  from files using a special directive `- "!file /path/to/list"`
- use a `set` instead of a `list` in the blacklist/whitelist entries, which is much faster with larger sets and deduplicates automatically.
For example:
```
blacklist:
- "evil.com"
- "hacker.org"
- "virus.cc"
- "!file /tmp/black.txt"
- "!file /tmp/black2.txt"
```
You can mix between the list entry definition formats, or use either one.
Feedback is welcome :)

please find it here:
https://github.com/SherifEldeeb/elastalert/commit/f3df37a052e55f246e9f402bd5a64ecd866969a0
.
P.S: Github noob.

pushkarevds:
Thank you very much!

sathishdsgithub:
@pushkarevds  @SherifEldeeb 

I would like to use the above method to whitelist IP address and specific string. Currently, I have a cardinality rule and I would like to do whitelist on few strings and IP address by combining the rule as shown below. The rule runs without any error but it's not working as expected. Can someone assist me is this a right method to combine the rule for whitelist?

**Main Rule**

```
type: cardinality
filter:
 - terms:
     rule_name:
         -  whitelist-zone  --> Rule name for whitelisting zone
         -  whitelisti-ip  --> Rule name for whitelisting IP address
``` 
**whitelist-zone**

```
root@ubuntu:/tmp/elastalert# more whitelist-zone.yaml
# ElastAlert Rule Name

name: whitelist-zone

# ElasticSearch host and port details

es_host: 192.168.96.141

es_port: 9200

# ElasticSearch Index Name

index: graylog_0

# Timestamp added to overcome ES graylog parsing issue

timestamp_field: timestamp
timestamp_type: custom
[...]

PR #1007, opened 2017-04-11
Allow for "external files" in blacklist/whitelist

SherifEldeeb:
This will do two things:
- Add a new option to include entries  from files using a special directive `- "!file /path/to/list"`
- use a `set` instead of a `list` in the blacklist/whitelist entries, which is much faster with larger sets and deduplicates automatically.
For example:
```
blacklist:
- "evil.com"
- "hacker.org"
- "virus.cc"
- "!file /tmp/black.txt"
- "!file /tmp/black2.txt"
```
You can mix between the list entry definition formats, or use either one.
.
fixes https://github.com/Yelp/elastalert/issues/711


SherifEldeeb:
fixes https://github.com/Yelp/elastalert/issues/711

Qmando:
Can you move expand_entries into __init__? It's opening the file for every single document.

SherifEldeeb:
Done (hopefully without breaking something else).
.
It is WAY faster now also :) 
Thanks for pointing out why.

Qmando:
Looks good. One last request: Can you add relevant documentation? docs/source/ruletypes.rst should be the correct file. Thanks!

SherifEldeeb:
Actually, thanks to you :)
Please inform me if there's anything else that needs to be done.

PR #1020, opened 2017-04-15
rule names made unique so errors not thrown when running against default folder

fabianlee:
On a out-of-the-box installation, invoking:

python -m elastalert.elastalert

Caused the following error because the 'name' field of two rules files had the same value. 

elastalert.util.EAException: Error loading file example_rules/example_new_term.yaml: Duplicate rule named Example rule

### DIFF
--- a/example_rules/example_frequency.yaml
+++ b/example_rules/example_frequency.yaml
@@ -17,7 +17,7 @@
 
 # (Required)
 # Rule name, must be unique
-name: Example rule
+name: Example frequency rule
 
 # (Required)
 # Type of alert.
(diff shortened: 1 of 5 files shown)


fabianlee:
Some individual rule files had es_host/es_port defined, and others relied on inheritance from main config.

A user would expect they could change the es_host in main config, and all the example rules would inherit these values (python -m elastalert.elastalert).

As it sits, when you run the most basic command (python -m elastalert.elastalert), you get lots of ElasticSearch cannot connect errors because half the rules use the main config's es_host, and the 3 rules that define it explicitly have it set to 'localhost'.

Qmando:
Thanks!
The examples are in dire need of updating. They haven't changed much in two years. My fault.

Issue #616, opened 2016-07-06
Custom time format issue


ochanism:
I am using a time format [ yyyy-MM-dd HH:mm:ss.SSS ] for elasticsearch data.
Thus, I tried to configure time format with custom type like this.

-------------------------------------------------------------------------------------------
timestamp_field: timestamp
timestamp_type: custom
timestamp_format: "%Y-%m-%d %H:%M:%S.%f"
-------------------------------------------------------------------------------------------

But, %f is for microsecond, not millisecond.
So I got this error.

----------------------------------------------------------------------------------------
[{u'reason': u'failed to parse date field [2016-07-04 02:27:56.274273] with format [yyyy-MM-dd HH:mm:ss.SSS]', u'type': u'parse_exception'}]
----------------------------------------------------------------------------------------

Is there any way to configure custom time format for millisecond?

Qmando:
Unfortunately, Python doesn't support millisecond-only accurate with strftime.

I have not really tested this, but as a temporary fix, you can try replacing this line (https://github.com/Yelp/elastalert/blob/d5617c3c610b2bedefbb6476dd2cfd99137fb16c/elastalert/util.py#L142) with ``return ts[:-3]``
That should make it cut off the last 3 digits of the timestamp.

Going the other way, python will accept .274 or .274000 milliseconds as equivalent, so that should be ok.

PR #1022, opened 2017-04-16
Added support to configurable expr evaluation to refine the ts obtained using a custom format

davidecavestro:
I have datetime values in ES formatted with milliseconds: similar to issue #616 but further complicated by the _Zulu_ timezone identifier ('Z' fixed char) after millis (i.e. _2017-04-16T06:32:55.776Z_).
This is a problem because - as far as I know - in python there's no easy way to specify a datetime format with millis precision (it has microsecond precision).

So I've added to elastalert config the support for an optional python expression to be evaluated in order to refine the timestamp obtained from `timestamp_format`.

The expression obtained by the `timestamp_format_expr` config entry, and is considered only when `timestamp_type` is `custom`.
When the expression is set, it gets evaluated with a namespace containing `ts` and `dt` as variables where: 
* `dt` is the datetime instance to convert
* `ts` is the timestamp obtained converting the datetime using `timestamp_format`
The evaluation result is returned.

So in my case I've set 
```
timestamp_type: custom
timestamp_format: '%Y-%m-%dT%H:%M:%S.%fZ'
timestamp_format_expr: 'ts[:23] + ts[26:]'
```

### DIFF
--- a/docs/source/ruletypes.rst
+++ b/docs/source/ruletypes.rst
@@ -534,12 +534,12 @@ timestamp_format_expr
 ^^^^^^^^^^^^^^^^^^^^^
 
 ``timestamp_format_expr``: In case Elasticsearch used custom date format for date type field, this option provides a way to adapt the
-value  obtained converting a datetime with ``timestamp_format``, when the format cannot match perfectly the format defined in Elastisearch.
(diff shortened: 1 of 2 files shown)


Qmando:
Can you add relevant documentation?

davidecavestro:
@Qmando I just documented the `timestamp_format_expr` option and its aim.

PR #1027, opened 2017-04-17
Fix an attributeerror on elasticsearch objects in some versions of th…

Qmando:
Fixes an error occurring with older versions of elasticsearch library introduced in #1006
```
self.current_es_addr = (self.current_es.host, self.current_es.port)
AttributeError: 'Elasticsearch' object has no attribute 'host'
```

These settings are applied to the global config before schema check, which is then used for the default in rules.


micahhausler:
Sorry for the bug! thanks!

PR #1030, opened 2017-04-17
fix KeyError: 'filtered' by passing 'five' flag

artushin:
Fixes what seems to be flag that forgot to be passed to `get_terms_query`.

### DIFF
--- a/elastalert/elastalert.py
+++ b/elastalert/elastalert.py
@@ -377,7 +377,7 @@ class ElastAlerter():
         base_query = self.get_query(rule_filter, starttime, endtime, timestamp_field=rule['timestamp_field'], sort=False, to_ts_func=rule['dt_to_ts'], five=rule['five'])
         if size is None:
             size = rule.get('terms_size', 50)
-        query = self.get_terms_query(base_query, size, key)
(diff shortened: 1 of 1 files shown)


artushin:
Stack trace from above error:
```
ERROR:root:Traceback (most recent call last):
  File "/opt/elastalert/elastalert/elastalert.py", line 914, in run_all_rules
    num_matches = self.run_rule(rule, endtime, self.starttime)
  File "/opt/elastalert/elastalert/elastalert.py", line 672, in run_rule
    if not self.run_query(rule, rule['starttime'], tmp_endtime):
  File "/opt/elastalert/elastalert/elastalert.py", line 474, in run_query
    data = self.get_hits_terms(rule, start, end, index, rule['query_key'])
  File "/opt/elastalert/elastalert/elastalert.py", line 380, in get_hits_terms
    query = self.get_terms_query(base_query, size, key)
  File "/opt/elastalert/elastalert/elastalert.py", line 195, in get_terms_query
    query_element['filtered'].update({'aggs': {'counts': {'terms': {'field': field, 'size': size}}}})
KeyError: 'filtered'
```

PR #1065, opened 2017-04-30
Dynamic PD Incident Key support

### LINKED ISSUE #765: Allow dynamic pagerduty_incident_key
This is especially helpful when generating pagerduty events for alerts that set query_key.

muz:
Akin to how `alert_text` and `alert_subject` accepts custom formatting based on values in the rule/alert; do the same for Incident Keys when posting to PD. This updates (and corrects!) the PDAlerter docs too.

This PR also fixes the Docker-Compose stuff, which I couldn't get to run prior to my changes.

I added a basic PDAlerter test to the test suite, and extended it with something that demonstrates the three valid use cases of incident keys (not specifying one, specifying an unformatted one, and a formatted one).

Fixes #765

### DIFF
--- a/Dockerfile-test
+++ b/Dockerfile-test
@@ -1,7 +1,7 @@
 FROM ubuntu:latest
 
 RUN apt-get update && apt-get upgrade -y
-RUN apt-get -y install python-setuptools python2.7 git tox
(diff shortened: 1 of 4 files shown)


bean5:
Awesome!

Issue #1085, opened 2017-05-11
Preserve lower-level stack traces when raising exceptions


danielpops:
Basically, do [this](http://stackoverflow.com/a/1350981/542442) any time an exception is raised after catching another one.

PR #1091, opened 2017-05-11
Added traceback info to EAExceptions

Qmando:
Add traceback info to places where we re-raise exceptions where the context may be important to debug


danielpops:
This addresses #1085

PR #1095, opened 2017-05-12
Changed the default ordering of alerts

Qmando:
This will make alerters default to going to the end versus the beginning. We set default for jira and email:
```
alerts_order = {
    'jira': 0,
    'email': 1
}
```
And this will make sure that jira (which uses the pipeline) will always come before something unless it explicitly has -1.

### DIFF
--- a/elastalert/config.py
+++ b/elastalert/config.py
@@ -372,7 +372,7 @@ def load_alerts(rule, alert_field):
             alert_field = [alert_field]
 
         alert_field = [normalize_config(x) for x in alert_field]
-        alert_field = sorted(alert_field, key=lambda (a, b): alerts_order.get(a, -1))
+        alert_field = sorted(alert_field, key=lambda (a, b): alerts_order.get(a, 1))
         # Convert all alerts into Alerter objects
         alert_field = [create_alert(a, b) for a, b in alert_field]
 


elastalert/config.py
```diff
@@ -372,7 +372,7 @@ def create_alert(alert, alert_config):
             alert_field = [alert_field]
 
         alert_field = [normalize_config(x) for x in alert_field]
-        alert_field = sorted(alert_field, key=lambda (a, b): alerts_order.get(a, -1))
+        alert_field = sorted(alert_field, key=lambda (a, b): alerts_order.get(a, 1))
```
danielpops: This is probably fine, unless you want to ensure that email always immediately follows JIRA, in which case you should set the default to something higher, like `alerts_order.get(a,2)`

danielpops: Maybe consider a future enhancement to make alert ordering semantics configurable within a rule?

muravitskiy: why do we need alert order map? `rule['alert']` is [always list](https://github.com/Yelp/elastalert/blob/5cdf3b3e409016c14e5bf76a0203729e9795fd5a/elastalert/config.py#L372) so valid order is already defined in rule config and pipelining should also work, just place `email` below `jira`
```yaml
alert:
   - "jira"
   - "email"

Qmando: So that people can take advantage of the alert pipeline without having to explicitly order the alerts correctly. Anything that adds something (like JIRA) can default to be run first.

PR #1096, opened 2017-05-14
Fix spike rules with alert_on_new_data and query_key

Hexta:
Fix false alerting on terms that have the same query key that previously matched
terms with enabled alert_on_new_data.

Changed windows moving algorithm. Do not clean current window for query key on
matched terms.
This change would also help to handle spike that was occurred right after the
previous one.


Hexta:
ping

Hexta:
ping

Issue #1078, opened 2017-05-05
Endtime of query windows of buckets slightly increasing


mrchang7:
I'm using 5s for run_every and 1s for bucket_interval in the metric_aggregation rule. Below are some EA output and I'm wondering why the end-time of the query window is slightly increasing. 

INFO:elastalert:Ran CPU_User from 2017-05-05 00:23:41.516443 EDT to 2017-05-05 00:23:46.569775 EDT: 10 query hits (0 already seen), 5 matches, 0 alerts sent
INFO:elastalert:Sleeping for 4.97194 seconds

INFO:elastalert:Ran CPU_User from 2017-05-05 00:23:46.516443 EDT to 2017-05-05 00:23:51.574753 EDT: 10 query hits (0 already seen), 5 matches, 0 alerts sent
INFO:elastalert:Sleeping for 4.972751 seconds

INFO:elastalert:Ran CPU_User from 2017-05-05 00:23:51.516443 EDT to 2017-05-05 00:23:56.579921 EDT: 10 query hits (0 already seen), 5 matches, 0 alerts sent
INFO:elastalert:Sleeping for 4.97243 seconds

INFO:elastalert:Ran CPU_User from 2017-05-05 00:23:56.516443 EDT to 2017-05-05 00:24:01.584863 EDT: 10 query hits (0 already seen), 5 matches, 0 alerts sent
INFO:elastalert:Sleeping for 4.972708 seconds

Qmando:
The end time is the current time when the query is generated. The 5 second interval is a just a best attempt by elastalert, it's not going to be exact. If some piece of code runs that takes a long time, it can be even further off.

Do you have a use case where exact intervals are required?

mrchang7:
Thanks for the answer. 

As for the use case, wondering if it's related to this issue I asked before: https://github.com/Yelp/elastalert/issues/1070

If I use bucket_interval 1s with the above configuration, sometimes (not always) it makes 6 queries with the last one having no hits - that could generate a false match in our use case.  

FYI, we don't want to use realert as we need state-based raise and clear of the alerts.

mircopolo:
Might the sync_bucket_interval setting be of any use here it should round the buckets to the nearest second and eliminate tiny precision errors.

mircopolo:
I actually thing I spotted a bug today, the code that is supposed to align the buckets with the start and end query times isn't working at all. This would definitely explain the partial buckets and probably the time drift as well. I'll submit a fix over the next few days.

PR #1116, opened 2017-05-22
Fix bucket offset calculation

mircopolo:
When bucket_interval is used (and sync_bucket_interval is false), ensure that the buckets keys align correctly with the query start and endtime from elastalert. Caused by a late rule key name changed in PR not being made in all the correct places.


mircopolo:
#1078

PR #1130, opened 2017-05-28
Add support for sending notifications to Microsoft Teams channels

JamesJJ:
Add support for sending notifications to Microsoft Teams channels.

20170529-ms-teams


Qmando:
LG2M!

PR #1145, opened 2017-06-05
add support for Microsoft Teams

sleongkoan:


elastalert/alerts.py
```diff
@@ -1295,3 +1295,36 @@ def alert(self, matches):
...
+        if isinstance(msteams_webhook_url, basestring):
+            msteams_webhook_url = [msteams_webhook_url]
+        self.msteams_webhook_url = msteams_webhook_url
+        self.msteams_proxy = self.rule.get('msteams_proxy')
+
+
+    def alert(self, matches):
+        payload = {
+            'isMultiline': True,
+            'text': '\n'.join(['```{}```'.format(match) for match in matches])
```
Qmando: Almost every other alerter uses ``BasicMatchString(self.rule, match)`` to generate the alert text. It's fine to have the raw matches too, but I'd like to see that behind a config setting.

I know the other alerters are very inconsistent, but those need to be fixed, not copied.

Qmando:
Can you also add some documentation?

elastalert/alerts.py
sleongkoan: Thanks for pointing that out. I now use `create_alert_body` to generate the body.

Documentation has also been updated

sleongkoan:
@Qmando I think this PR is better than mine: https://github.com/Yelp/elastalert/pull/1130
you should have a look

sleongkoan:
#1130 is better than this, closing this down

PR #1151, opened 2017-06-06
Add environment variable substitution to use_kibana4_dashboard

### LINKED ISSUE #1150: Use environment variables in rule files
Hi there,

Thanks for ElastAlert, it's pretty cool :-)

I'm using ElastAlert inside a Docker container and I'd like to be able to tweak some settings based on environment variables.

For example, when using Docker Machine on Windows, the Kibana link I send in reports need to point to the Kibana host, whose real IP is the result of `docker-machine ip` and varies from one machine to the other:

```yaml
use_kibana4_dashboard: "http://192.168.99.100:7007/#/discover/My-Saved-Search"
```

I also intend on deploying these same alerting rules to one of our production clusters in which the Kibana host will obviously not be running in my local VM ;-)

I'm willing to put some time into a PR if you're willing to accept the contribution!

André

AndreLouisCaron:
Without this, the Kibana host is hardcoded in the rule definition and some kind
of external rule templating is required to share the rule across multiple
deployements.  This change allows doing the same without a templating engine:
all we need is for the caller to point environment variables of their choice to
the right location when running ElastAlert.

Fixes #1150

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,13 @@
 # Change Log
 
+
+## Unreleased
+
+### Added
+- Added the ability to expand environment variables in
+  ``use_kibana4_dashboard``, enabling sharing of rules across multiple
+  deployments of ElastAlert and Kibana.
+
 ## v0.1.11 - 2017-04-19
 
 ### Fixed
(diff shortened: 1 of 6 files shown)


AndreLouisCaron:
Thanks for merging :-)

PR #1163, opened 2017-06-12
Added support for lists of emails in email_from_field

### LINKED ISSUE #1064: email_from_field as an array?
I'm trying to configure elastalert to read an array from data which looks like:

```
"orgManagers": [
    "test@test.com",
    "a@b.com"
], ...
```

And my alert configuration looks like:

`email_from_field: "orgManagers"`

Now this works as a single email address, but I need to send the alert to multiple recipients not known by the alert itself...those recipients come from the data. I've looked at enhancements but it looks like only the email field supports multiple. Ideas?

Qmando:
Fixes #1064

### DIFF
--- a/elastalert/alerts.py
+++ b/elastalert/alerts.py
@@ -398,6 +398,8 @@ class EmailAlerter(Alerter):
                     to_addr = [recipient + self.rule['email_add_domain']]
             elif isinstance(recipient, list):
                 to_addr = recipient
+                if 'email_add_domain' in self.rule:
+                    to_addr = [name + self.rule['email_add_domain'] for name in to_addr]
         email_msg = MIMEText(body.encode('UTF-8'), _charset='UTF-8')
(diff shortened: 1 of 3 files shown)


elastalert/alerts.py
```diff
@@ -396,6 +396,8 @@ def alert(self, matches):
                     to_addr = [recipient]
                 elif 'email_add_domain' in self.rule:
                     to_addr = [recipient + self.rule['email_add_domain']]
+            elif isinstance(recipient, list):
+                to_addr = recipient
```
danielpops: Do you need/want similar logic as above? e.g. could I specify:

```
email_from_field:
- dpopes
- qlo
email_add_domain: "@yelp.com"
```

y-trobinso:
lgtm. To be pedantic, using BCC might be more correct here, but this otherwise looks correct to me.

PR #1166, opened 2017-06-13
twilio_accout_sid renaming to 'account'?

djmrr:
- Currently using Elastalert and loving it
- Just figured that maybe there might have been a typo in the naming?
- Kinda thought "account" sounds / seems better than "accout"

Open to all input, thoughts, questions, etc!


Qmando:
Good catch. At least it was consistent everywhere ;)

PR #1171, opened 2017-06-16
Add ca_certs param

alesnav:
Add ca_certs param to util.py. This way, you can specify your custom CA root to use with verify_certs when it is a self-signed one.


alesnav:
Thank you @Qmando 👍

PR #1179, opened 2017-06-21
fix a bug where realert doesnt respect query_key when using flatline

### LINKED ISSUE #1177: Rule with flatline and use_terms_query doesn't work anymore after elastalert upgrade (->v0.1.12)
Hello everybody,

ES version : 2.4.2

After have upgraded from elastalert  **v0.0.93** to elastalert **v0.1.12**, the following alert fail :

```
name: "##System_Log_Collection_Stopped"
index: graylog_*
type: flatline
query_key: event_src
use_terms_query: true
doc_type: message
threshold: 1
timeframe:
  minutes: 3

filter:
- query:
    query_string:
      query: "is_system_device:true"

realert:
    minutes: 5
```
 
When the first event_src stop to send logs, an alert is raised. 
But , when a another event_src stop to send log (i.e five minutes later), no alert is raised. 

The buckets (two for my test) are well seen by elastalert. elastalert see the two buckets at the beginning, then he notice the first missing bucket, and finally the second (two bucket -> one bucket -> zero bucket)

Now, the strange thing :  I tested the same alert **on elastalert v0.0.93 (ES version is also 2.4.2) and it works fine** : one alert when the first event_src stop to send logs and another one when the second event_src stop to send logs.

With elastalert **v0.0.93** ,it seems that realert is not working as I don't have any other alert than the first two alerts . 

With elastalert **v0.1.12**, it seems that realert is working well as I have new alerts every five minutes 

So, I would suspect a bug in **v0.1.12** around flatline and use_temrs_query but I'm asking if somebody else has noticed the same behavior as the issue could be due to my own specific environment 

(without use_terms_query, alert is fine but I have to use use_terms_query for performance reasons)

Thanks for your feedback ! :-)

Qmando:
Fixes #1177 

This is a hacky fix, but I don't want to break backwards compatibility, at least not yet. Flatline + query_key has been using this "key" field forever.

### DIFF
--- a/elastalert/elastalert.py
+++ b/elastalert/elastalert.py
@@ -25,6 +25,7 @@ from croniter import croniter
 from elasticsearch.exceptions import ElasticsearchException
 from elasticsearch.exceptions import TransportError
 from enhancements import DropMatchException
+from ruletypes import FlatlineRule
 from util import add_raw_postfix
 from util import cronite_datetime_to_timestamp
 from util import dt_to_ts
@@ -686,6 +687,9 @@ class ElastAlerter():
 
     def get_query_key_value(self, rule, match):
         # get the value for the match's query_key (or none) to form the key used for the silence_cache.
+        # Flatline ruletype sets "key" instead of the actual query_key
+        if isinstance(FlatlineRule, rule['type']) and 'key' in match:
+            return match['key']
         return self.get_named_key_value(rule, match, 'query_key')
 
     def get_aggregation_key_value(self, rule, match):


elastalert/elastalert.py
```diff
@@ -686,6 +687,9 @@ def get_segment_size(self, rule):
 
     def get_query_key_value(self, rule, match):
         # get the value for the match's query_key (or none) to form the key used for the silence_cache.
+        # Flatline ruletype sets "key" instead of the actual query_key
+        if isinstance(rule['type'], FlatlineRule) and 'key' in match:
+            return match['key']
```
danielpops: I think you should still go through `get_named_key_value`, but pass `key` instead of `query_key`, so that you can get the benefits that come with it (unicode stuff, using `lookup_es_key` for better handling of complex keys)

Qmando: get_named_key_value won't work for this. It takes a key name then looks up the value of that key in the rule. Passing 'query_key' might return rule['username'] for example. I will add a unicode( ) though.

PR #1191, opened 2017-06-28
adding ability to define multiple import files per rule file

eladamitpxi:


eladamitpxi:
any chance of getting this merged?

sathishdsgithub:
@eladamitpxi 

Could you please help me to understand in which scenario I can use the above "multiple import file" options.

Presently I'm working on the below scenario in which I have two different `.yaml `file and each file has its own unique import rule name. However, this is not working as expected . In the current elastalert configuration we can only import one file at the moment. Though, that file can also import things.

The imported file will also overwrite any options in the first rule, so we can't have two whitelists and make them merge together.

If I use the above "ability to define multiple import files" will it work in my scenario? Appreciate your help

```


../SIEM-tests/whitelist-outbreak.yaml
::::::::::::::

import: outbreak-scan

type: whitelist

compare_key: srx-source-address

ignore_null: true

whitelist:

  - "!file /opt/whitelist-oubreak.txt"
::::::::::::::
../SIEM-tests/whitelist-zone.yaml
::::::::::::::

name: port scan rule

import: port-scan

type: whitelist

compare_key: srx-source-address


ignore_null: true

whitelist:
   - "!file /opt/whitelist-ip.txt"

```

eladamitpxi:
the scenario we added this for is more around re-use of configs rather than composition of rules
e.g. imaging the following files:
filters/foo_filter -> a filter
filters/bar_filter -> another filter
alerts/production-slack
alerts/opsgenie-team-a
alerts/opsgenie-team-b

and a rule which does the following:
...
import:
  - filters/foo_filter
  - alerts/opsgenie-team-a
...

and another which does:
...
import:
  - filters/bar_filter
  - alerts/opsgenie-team-a
...

sathishdsgithub:
@eladamitpxi 

So, basically, you can import two rule files in a single `.yaml `rule like the below. Correct me if I'm wrong 

**main-rule.yaml**

import: 
 - rul1.yaml

import:
- rule2.yaml

In **rule1.yaml**

```
whitelist:
 - "10.0.0.1"
 - "1.2.3.4"
```

**rul2.yaml**

```
blacklist:
 - "1.1.1.1"
- "2.2.2.2"
```

eladamitpxi:
yes, but the type attribute will get overwritten and only one will be used

sathishdsgithub:
@eladamitpxi 

Thanks 

I tried to create two different  whitelist .yaml rule and each whitelist rule has unique import rule name. But the first rule always overwrites the second one. Is there any options to overcome this ?

eladamitpxi:
not with the current implementation


Thanks,
Elad Amit | Architect
Mobile: +972-54-6939204
PerimeterX

sathishdsgithub:
@eladamitpxi 

Can you share some example for import filters to the rule? 

For example, i have the below filter in the name of `sample-rule.yaml`. 

```
filter:

 - not:
       terms:
         srx-source-address:   ["10.18.109.59", "10.10.10.10", "20.20.20.20", "30.30.30.30"]

```

How do I import the above filter to a different rule?. I tried with below options it's not working 

```
filter:
- and:
      filters: [sample-rule.yaml]
```

http://elastalert.readthedocs.io/en/latest/recipes/writing_filters.html?highlight=filters%20from%20rule.yaml#writing-filters-for-rules

eladamitpxi:
imported filters are merged with ANDs so i'm not sure the and there is
required
see https://github.com/Yelp/elastalert/blob/master/elastalert/config.py#L126


Thanks,
Elad Amit | Architect
Mobile: +972-54-6939204
PerimeterX

eladamitpxi:
@Qmando @danielpops any chance of getting this merged?

pataquets:
@eladamitpxi : Thanks for the work! Definitely need this. We would like to separate alerting configurations by subsystem and teams, with different Slack channels and webhook URLs. We would like also to import secrets from standalone, unversioned (read: gitignore'd) files.
@Qmando @danielpops : Anything else needed to do here? I would happily help with testing, if there is a chance for this to move forward.

nsano-rururu:

eladamitpxi:
this PR is no longer relevant due to the drift in code, an updated one can be seen here https://github.com/Yelp/elastalert/pull/3095

PR #1201, opened 2017-07-05
Set shards and replicas settings for index creation

mjflve:
Hi,

This PR is intended to give the option to set shards and replicas for the elastalert index creation, this could be useful in environments with one elasticsearch node.

Best Regards

### DIFF
--- a/config.yaml.example
+++ b/config.yaml.example
@@ -55,3 +55,8 @@ writeback_index: elastalert_status
 # sending the alert until this time period has elapsed
 alert_time_limit:
   days: 2
+
+# Set shards and replicas settings for index creation
+#index_settings:
+#  shards: 1
+#  replicas: 0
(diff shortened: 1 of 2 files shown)


eht16:
For anyone interested, here is an updated diff: https://gist.github.com/eht16/135e3e3e21d5a0506a909febf574fd07

I think this would be a useful addition. For example, my use case is also a one-node ES cluster so I don't need replicas and for rather small indices I prefer only one or two shards.

admlko:
I agree, this would be nice to have. In my case, there is no need for extra 4 primary shards taking up resources.

@eht16 Cheers for the diff :)

benetis:
Any reason why this wasn't merged?

mmguero:
I've created a new updated diff based on the one @eht16 created on Mar 21, 2018.

If interested, [it's here](https://gist.githubusercontent.com/mmguero/3dde45220d0482a44873b493b46c47ba/raw/a6254b8d84ac325fde290ae5e32d94da2b503536/create_index.py.diff).

nsano-rururu:

PR #1203, opened 2017-07-07
Wait for provisioning

### LINKED ISSUE #1199: Fails to start when provisioned with docker-compose
Hi there!

I have a local development setup with `elastalert` running inside a Docker container.  However, I'm having trouble using `docker-compose up -d` to provision my stack: the `elastalert` container exits almost immediately because ElasticSearch is not yet responsive and/or the `elastalert_status` index does not yet exist.

To work around this issue, I need to do something like this:

```
$ docker-compose up -d elasticsearch ...  # everything except elastalert
$ # Wait until ElasticSearch is responsive (e.g. poll using cURL)
$ elastalert-create-index ...
$ docker-compose up -d elastalert
```

For larger configurations, this quickly becomes tedious because `docker-compose` doesn't supply a way to exclude a single service; instead. you have to list every single service except for the one you don't want to start.

I wish I could start `elastalert` immediately and that it would internally poll ElasticSearch until it was ready and then proceed as usual.  This would allow me to simply:

```
$ docker-compose up -d
$ elastalert-create-index ...
```

Any chance you would be willing to accept such a contribution?

AndreLouisCaron:
Fixes #1199.

In addition to the new unit tests, I tried testing this manually.  I'm having a hard time building my Docker container against a custom branch, but it works in well in foreground mode (e.g. run in my terminal against an ElasticSearch node inside a Docker container).

It will respect the delay and bail out and you can always CTRL-C/SIGINT at any time if you lose patience before the timeout elapses.

The one thing I find unfortunate is that `elasticsearch-py` is very verbose on connection errors, so the polling spams with exceptions tracebacks until the connection succeeds (there are no tracebacks if it succeeds on the first attempt).  I don't know any way to make this more pleasing without changing the logging inside `elasticsearch-py` itself.

### DIFF
--- a/elastalert/elastalert.py
+++ b/elastalert/elastalert.py
@@ -7,6 +7,7 @@ import os
 import signal
 import sys
 import time
+import timeit
 import traceback
 from email.mime.text import MIMEText
 from smtplib import SMTP
@@ -22,6 +23,7 @@ from config import get_rule_hashes
 from config import load_configuration
 from config import load_rules
 from croniter import croniter
+from elasticsearch.exceptions import ConnectionError
 from elasticsearch.exceptions import ElasticsearchException
 from elasticsearch.exceptions import TransportError
 from enhancements import DropMatchException
@@ -36,6 +38,7 @@ from util import elasticsearch_client
 from util import format_index
 from util import lookup_es_key
 from util import parse_deadline
+from util import parse_duration
 from util import pretty_ts
 from util import replace_dots_in_field_names
 from util import seconds
@@ -76,6 +79,11 @@ class ElastAlerter():
                                                           'Use "NOW" to start from current time. (Default: present)')
         parser.add_argument('--end', dest='end', help='YYYY-MM-DDTHH:MM:SS Query to this timestamp. (Default: present)')
         parser.add_argument('--verbose', action='store_true', dest='verbose', help='Increase verbosity without suppressing alerts')
+        parser.add_argument('--patience', action='store', dest='timeout',
+                            type=parse_duration,
+                            default=datetime.timedelta(seconds=30),
+                            help='Maximum time to wait for ElasticSearch to become responsive.  Usage: '
+                            '--patience <units>=<number>. e.g. --patience minutes=5')
         parser.add_argument(
             '--pin_rules',
             action='store_true',
@@ -980,6 +988,7 @@ class ElastAlerter():
                 except (TypeError, ValueError):
                     self.handle_error("%s is not a valid ISO8601 timestamp (YYYY-MM-DDTHH:MM:SS+XX:00)" % (self.starttime))
                     exit(1)
+        self.wait_until_responsive(timeout=self.args.timeout)
         self.running = True
         elastalert_logger.info("Starting up")
         while self.running:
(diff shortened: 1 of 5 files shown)


elastalert/elastalert.py
```diff
@@ -994,6 +1010,25 @@ def start(self):
...
+
+        # Elapsed time is a floating point number of seconds.
+        timeout = timeout.total_seconds()
+
+        # Periodically poll ElasticSearch.  Keep going until ElasticSearch is
+        # responsive *and* the writeback index exists.
+        ref = clock()
+        while (timeout is None) or ((clock() - ref) < timeout):
+            try:
+                if self.writeback_es.indices.exists(self.writeback_index):
```
Qmando: Can we add

```
else:
    logging.error('ElastAlert status index "{}" does not exist. '
                           'Please run elastalert-create-index'.format(self.writeback_index))
    exit(1)
```

or something similar?

I don't want someone to get stuck in an infinite loop if they have failed to create the index or have a typo in their config.

AndreLouisCaron: Sure, sounds good.  Will update ASAP.

AndreLouisCaron: BTW, there is currently not possibility to end up with `timeout is None` so I will remove it (there will never be an infinite loop).

elastalert/elastalert.py
```diff
@@ -994,6 +1010,36 @@ def start(self):
...
+
+        # Elapsed time is a floating point number of seconds.
+        timeout = timeout.total_seconds()
+
+        # Periodically poll ElasticSearch.  Keep going until ElasticSearch is
+        # responsive *and* the writeback index exists.
+        ref = clock()
+        while (clock() - ref) < timeout:
+            try:
+                if self.writeback_es.indices.exists(self.writeback_index):
```
Qmando: Can you put an
```
else:
    break
```
here?

It seems that if the index doesn't exist, this will still spin until the timeout period ends.

AndreLouisCaron: Yeah, that's intentional.

I understand that the typical use case involves setting up an ElasticSearch cluster, then running `elastalert-create-index`, but as described in the issue, I run `elastalert-create-index` after  `docker-compose up -d` that spawns both ElasticSearch & ElastAlert.

In order to make that work, I need this loop to wait until both these conditions are true:

1. ElasticSearch is responsive; and
2. The writeback index exists.

Qmando: I see. That's very reasonable. 

I would like to lower the default timeout in that case. If someone had incorrectly configured the index, I expect that they would be confused by the lack of output and might terminate the program before the timeout expired.

Maybe something like 1 second, which would allow the initial check to happen but would not even mislead people as to why nothing was happening. Thoughts?

AndreLouisCaron: I'm totally fine with lowering the default.  It can even be 0 so that we preserve the current behavior by default.

Qmando: 0 or 1 is fine with me.

0 would actually cause the current implementation to break (loop never enters so it would eventually exit 1), so I guess that should be fixed either way as someone might attempt to use 0.

AndreLouisCaron: Yeah, I was thinking the same.  I will add a special case for `timeout <= 0.0` to avoid entering the lop.

Qmando:
Looks good to me!

AndreLouisCaron:
@Qmando Thanks for the help and for your patience :-)

PR #1205, opened 2017-07-07
Add jira_bump_after_inactivity field

jchuah:
- Only bumps ticket after jira_bump_after_inactivity of inactivity
- Add arrow to requirements

### DIFF
--- a/elastalert/alerts.py
+++ b/elastalert/alerts.py
@@ -14,6 +14,7 @@ from smtplib import SMTPAuthenticationError
 from smtplib import SMTPException
 from socket import error
 
+import arrow
 import boto3
 import requests
 import stomp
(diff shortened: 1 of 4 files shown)


elastalert/alerts.py
```diff
@@ -687,6 +690,12 @@ def alert(self, matches):
         if self.bump_tickets:
             ticket = self.find_existing_ticket(matches)
             if ticket:
+                inactivity_datetime = arrow.now().replace(days=-self.bump_after_inactivity)
```
Qmando: I assume this ticket.fields.updated is a ISO8601 string?
To keep consistent with the rest of the code, can you use util.ts_now and util.ts_to_dt
```
from util import ts_now, ts_to_dt
...
inactivity_datetime = ts_now() - datetime.timedleta(days=self.bump_after_inactivity)
if ts_to_dt(ticket.fields.updated) ...
```

jchuah: agreed, thanks!

PR #1207, opened 2017-07-08
Feature/advanced post alerter

brianmpollack:
Added alerter to send JSON to generic HTTP POST endpoint.

### DIFF
--- a/docs/source/ruletypes.rst
+++ b/docs/source/ruletypes.rst
@@ -1625,23 +1625,23 @@ This alert type will send results to a JSON endpoint using HTTP POST. The key na
 
 Required:
 
-``advanced_post_url``: The URL to POST.
-
-``advanced_post_payload``: List of keys:values to use as the content of the POST. Example - ip:clientip will map the value from the clientip index of Elasticsearch to JSON key named ip.
+``http_post_url``: The URL to POST.
 
 Optional:
 
-``advanced_post_static_payload``: Key:value pairs of static parameters to be sent, along with the Elasticsearch results. Put your authentication or other information here.
+``http_post_payload``: List of keys:values to use as the content of the POST. Example - ip:clientip will map the value from the clientip index of Elasticsearch to JSON key named ip. If not defined, all the Elasticsearch keys will be sent.
 
-``advanced_post_proxy``: URL of proxy, if required.
+``http_post_static_payload``: Key:value pairs of static parameters to be sent, along with the Elasticsearch results. Put your authentication or other information here.
+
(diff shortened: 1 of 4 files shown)


Qmando:
How does this differ from https://github.com/Yelp/elastalert/blob/master/elastalert/alerts.py#L1326?

brianmpollack:
My new post alerter allows the user to define the POST key for each Elasticsearch key. For example, in Elasticsearch there might be a key called "clientip" but you want the key when it gets sent to the endpoint to be "ip". The user can also define static values that get sent to the endpoint, such as their authentication credentials.

The reason I added this is that I have endpoints which I cannot customize, but want to POST data to them. I had to conform to their key/value structure.

I agree. Would you prefer that I incorporated this into the "Simple" post alerter and just call it "PostAlerter"? I can then add documentation for the alerter as a whole.

Qmando:
I'd prefer they be merged into a single alerter. You can drop the word "Simple" 😄

brianmpollack:
Thanks! Sounds good. I combined them, updated the names, and updated the documentation.

As a brief summary, if you leave the http_post_payload field undefined, it will just send all the Elasticsearch keys as-is.

FraVieMor:
Hi!
I want do a HTTP Post to a different url (and payload) for each match. I think that this PR is useful for me. I was looking for an alert to get this behavior. Thanks!

Would be great if it is documented.

brianmpollack:
@FraVieMor I'm glad this will be useful to you! I have documentation in the docs/source/ruletypes file that will be automatically added to the online documentation if this PR gets approved.

av3ri:
@brianmpollack @Qmando this is awsesome! I got a chance to test the new alerter and it works really well. 😊 May I suggest a small enhancement? The ability to add header values to the post. I was able to do it in the alert.py file manually but it would be perfect if you could define them in the alert. E.g: "Authentication" for token based webhooks.

brianmpollack:
@av3ri I'm glad you like it! That's funny, I actually was thinking of that the other day. I may add that when I get the time. As a quick tip, if you're using basic HTTP authentication, you can do http(s)://username:password@example.com/foo/ rather than adding to the header.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication#Access_using_credentials_in_the_URL

av3ri:
@brianmpollack haha great minds think alike! 😊thanks! Yes that's definitely a handy tip! Looking forward to seeing how you go. Thanks again for your efforts with making the post alerter even better. I am currently using this to populate a webhook I built for HTML based email alerts. Where I have a lot more freedom to customise the email alert.

PR #1225, opened 2017-07-14
multi es-instances support

clarkchen:
I take advantages of Elasticsearch, make host param as a list.

Host Config Change to 
"host1:port1, host2:port " can handle multi es-instances

while  the original port and host can also work well here

### DIFF
--- a/elastalert/util.py
+++ b/elastalert/util.py
@@ -2,7 +2,6 @@
 import os
 import datetime
 import logging
-import json
 import dateutil.parser
 import dateutil.tz
 from auth import Auth
@@ -318,7 +317,6 @@ def build_es_conn_config(conf):
             es_port = int(os.environ.get('ES_PORT', conf['es_port']))
         parsed_conf['es_host'] = parse_host(es_host, es_port)
 
-
     if 'es_username' in conf:
         parsed_conf['es_username'] = os.environ.get('ES_USERNAME', conf['es_username'])
         parsed_conf['es_password'] = os.environ.get('ES_PASSWORD', conf['es_password'])
@@ -359,8 +357,6 @@ def parse_host(host, port="9200"):
         return host_list
     else:
         return ["{host}:{port}".format(host=host, port=port)]
-
-
 
 
 def parse_duration(value):
--- a/elastalert/util.py
+++ b/elastalert/util.py
@@ -2,7 +2,7 @@
 import os
 import datetime
 import logging
-
+import json
 import dateutil.parser
 import dateutil.tz
 from auth import Auth
@@ -283,7 +283,6 @@ def elasticsearch_client(conf):
                                      profile_name=es_conn_conf['profile'])
 
     return Elasticsearch(host=es_conn_conf['es_host'],
-                         port=es_conn_conf['es_port'],
                          url_prefix=es_conn_conf['es_url_prefix'],
                          use_ssl=es_conn_conf['use_ssl'],
                          verify_certs=es_conn_conf['verify_certs'],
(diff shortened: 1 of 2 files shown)


clarkchen:
hello?

elastalert/util.py
```diff
@@ -308,12 +306,17 @@ def build_es_conn_config(conf):
...
-    parsed_conf['es_host'] = os.environ.get('ES_HOST', conf['es_host'])
-    parsed_conf['es_port'] = int(os.environ.get('ES_PORT', conf['es_port']))
     parsed_conf['es_url_prefix'] = ''
     parsed_conf['es_conn_timeout'] = conf.get('es_conn_timeout', 20)
     parsed_conf['send_get_body_as'] = conf.get('es_send_get_body_as', 'GET')
 
+    if "es_host" in conf:
+        es_host = os.environ.get('ES_HOST', conf['es_host'])
+
+        if "es_host" in conf:
```
Qmando: es_port

Though, this also isn't necessary.

elastalert/util.py
```diff
@@ -308,12 +306,17 @@ def build_es_conn_config(conf):
     parsed_conf['es_password'] = None
     parsed_conf['aws_region'] = None
     parsed_conf['profile'] = None
-    parsed_conf['es_host'] = os.environ.get('ES_HOST', conf['es_host'])
-    parsed_conf['es_port'] = int(os.environ.get('ES_PORT', conf['es_port']))
     parsed_conf['es_url_prefix'] = ''
     parsed_conf['es_conn_timeout'] = conf.get('es_conn_timeout', 20)
     parsed_conf['send_get_body_as'] = conf.get('es_send_get_body_as', 'GET')
 
+    if "es_host" in conf:
```
Qmando: This check isn't necessary, it's always there.

elastalert/util.py
```diff
@@ -344,6 +347,18 @@ def build_es_conn_config(conf):
     return parsed_conf
 
 
+def parse_host(host, port="9200"):
```
Qmando: Why not accept a list instead of a comma separated string?

YoavBZ:
Nice! I've just started working on a similar feature :)
In my case, I wanted to run rules on few different hosts and didn't wanted to duplicates my rules, so I added a list of objects to the rule .yaml file that includes the ES hosts, ports and names, so the rule will run separately on each host.
I also appended the environment name to the rule name, i.e "Rule name [Env name]".

How this one is going to work?

Qmando:
I guess the motivation here is that you could replace a load balancer by having a pool of servers to roundrobin requests to.

YoavBZ:
I see.. My idea was more for multiple environments support

Qmando:
So you implemented this feature to avoid duplicating YAML files, is that correct?

You can import bits and pieces of rules, so you could specify two files with just a es_host, name and import parameter.

YoavBZ:
Yes, I wanted to avoid duplicating cause I have one rule with 2/3 ES hosts I want to run it on.
So instead of creating few duplicate rules but with different es_hosts I created one rule with few es_hosts in it

YoavBZ:
*Not one rule actually, I need to run all my rules on few es_hosts

sathishdsgithub:
@YoavBZ @Qmando @clarkchen 

Can we make use of the above modified `util.py` for multi ES instance? I tried with below format but it did not work.
Please, can someone suggest me the right method to mention the list of ES host entry with port?

`es_host: "192.168.96.157:9200, 192.168.96.141:9200"`

nsano-rururu:

clarkchen:
> Can you resolve conflicts please :)

sorry, I lost my origin project an start a new pull [request](https://github.com/Yelp/elastalert/pull/3109), where you can review code there

clarkchen:
@sathishdsgithub 

the `ElasticSearchClient` init should change to hosts
image


I add a test case which connect with a docker-composed elastic cluste

you can find the test case [here](https://github.com/clarkchen/elastalert/blob/181103918fb3b684b1b497140aa274d69bfb4a34/tests/util_test.py#L242)

hope it will work

PR #1267, opened 2017-08-06
Add match_time to alert

RubieV:
Currently, Kibana can only query alerts by their creation time, not the time of the event. The event time is stored in the match_body, however that is not indexed. 

This change adds the match_time field to an alert, and adds it as indexed field to the ElasticSearch index. In order to make use of the indexin feature, you will have to reindex.


RubieV:
@Qmando There is some great ElastAlert in Kibana update coming, would you be able to merge this into trunk?

PR #1286, opened 2017-08-14
adds eval function to jira alerter

ytonui:
This change enables the Jira alerter to load values dynamically from the match.


Qmando:
Looks good to me!

PR #1292, opened 2017-08-16
Fix bug in config processing of ES_USE_SSL in which it could never be false

### LINKED ISSUE #1291: env var ES_USE_SSL=false doesn't work
No matter what I set ES_USE_SSL env var to, it kept trying to run with ssl enabled. This is because the env var is a string, not a boolean, so if "false" is always true.

kellycampbell:
Fix bug in config processing of ES_USE_SSL in which it could never be false.
Also added username/password flags to create_index

Fixes #1291

### DIFF
--- a/elastalert/config.py
+++ b/elastalert/config.py
@@ -416,7 +416,6 @@ def load_rules(args):
 
     for env_var, conf_var in env_settings.items():
         val = env(env_var, None)
-        print("val[%s] = %s" % (env_var, val))
         if val is not None:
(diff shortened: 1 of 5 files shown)


Qmando:
Looks good to me. I'd rather you remove the setup.py version bump though, then I'll merge it.

kellycampbell:
Yup, I removed the couple of things that shouldn't be there after the first commit. Check it now.

Qmando:
Thanks for doing this!

PR #1293, opened 2017-08-16
Read logging configuration from config file

eht16:
Support reading a whole dict logging configuration from the config file
and configure the logging framework accordingly.
Move special logging configuration via command line options
(--verbose, --debug) into load_rules() as it needs to be set after the
logging framework has been reconfigured.

This should keep the previous behavior as close as possible when no
logging setup is provided in the configuration and mimic the desired log
level adjustments where appropriate even if a custom logging config
is given.

My main motivation was to add a custom log format, especially including
a timestamp (I don't like logs without timestamps).
Furthermore, with a full logging config, you can easily log your
ElastAlert logs into Logstash with https://github.com/eht16/python-logstash-async/, e.g. https://gist.github.com/eht16/b2183e1ce8373b8916b72dfa985a20c6.

### DIFF
--- a/config.yaml.example
+++ b/config.yaml.example
@@ -62,3 +62,53 @@ writeback_index: elastalert_status
 # sending the alert until this time period has elapsed
 alert_time_limit:
   days: 2
+
+# Custom logging configuration
+# If you want to setup your own logging configuration to log into
+# files as well or to Logstash and/or modify log levels, use
+# the configuration below and adjust to your needs.
+# Note: if you run ElastAlert with --verbose/--debug, the log level of
+# the "elastalert" logger is changed to INFO, if not already INFO/DEBUG.
+#logging:
+#  version: 1
+#  incremental: false
+#  disable_existing_loggers: false
+#  formatters:
+#    logline:
+#      format: '%(asctime)s %(levelname)+8s %(name)+20s %(message)s'
+#
+#    handlers:
+#      console:
+#        class: logging.StreamHandler
+#        formatter: logline
+#        level: DEBUG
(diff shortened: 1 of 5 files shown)


pataquets:
+1 if:
* Command-line switches prevail if they contradict config file, to avoid unexpected behaviour changes.
* Command-line switches are kept to avoid breaking anyone's setup.

Would love to get this in, since I would more easily enable debug-level logging (since ```--debug``` does not raise logger's level) and this will ease development & troubleshooting.

pataquets:
Also, I'm not that deep in Python as of now, but ideally ES tracing should be as configurable as main logging, too.

eht16:
I'd be happy to update the PR if there are any chances that it get merged eventually.
Of course, the current behaviour should not be changed, I tried to ensure this but will re-check to be sure.

eht16:
I rebased my changes on current master, so everything is up to date and "mergeable" :).

Also I tested again the code for changes in current behaviour regarding `--verbose` and `--debug`.

There is only one change but on purpose:
If the log level in config is set to `DEBUG` and `--verbose`/`--debug` do *not* alter it to INFO.
Strictly it should be altered to `INFO` as this is what the previous behaviour did but only because the log level was never set lower than `INFO`.

So I personally think this is the right thing to do as it would seem strange to me if I put `DEBUG` in the config and `--debug` on the CLI and the resulting log level
is `INFO`.


@Qmando (or whoever else) any change to get this merged?

eht16:
I rebased my changes on current master again, so everything is up to date and "mergeable" :).

eht16:
Thanks!

PR #1295, opened 2017-08-18
Logging the unique values in values list of NewTermsRule

byakuinss:
Hi, this PR is about New Term Rule type. 

According to this line in `ruletypes.py`, it should get the unique values count of `values` list, but it counts all values in the list. 
```
self.seen_values[key] = list(set(values))
elastalert_logger.info('Found %s unique values for %s' % (len(values), key))
```

I also added another line to print `values` list for debug 
```
elastalert_logger.info('Found %s unique values for %s' % (len(values), key))
elastalert_logger.info(values)
```
image

So I changed logging.info to print the number of distinct values in `values` list, and print `set(values)` for debug. 
```
elastalert_logger.info('Found %s unique values for %s' % (len(set(values)), key))
elastalert_logger.info(set(values))
```
image

### DIFF
--- a/elastalert/ruletypes.py
+++ b/elastalert/ruletypes.py
@@ -667,7 +667,7 @@ class NewTermsRule(RuleType):
                         elastalert_logger.info('Found no values for %s' % (field))
                     continue
                 self.seen_values[key] = list(set(values))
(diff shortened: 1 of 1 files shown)


Qmando:
Nice find!!

PR #1302, opened 2017-08-24
Correct invalid args attribute

gangefors:
Changing args.passowrd to args.password


gangefors:
Seems that there are no tests for the added options to `elastalert-create-index` script.

If you want me to add some just give me a mention.

PR #1303, opened 2017-08-24
Clarify CLI message

bean5:
This one has always been a little unclear and every time I start a new ElastAlert instance from scratch with new rules, I end up using debug/verbose modes and for some reason I end up thinking they can both run simultaneously. This clarifies the CLI message so I remember sooner.


Qmando:
Mind splitting this string into two lines?

It's tripping the PEP8 style checks.

bean5:
I can try. I'm still new to Python.

Qmando:
The usual way in Python is:
```
logger.info('Lorem ipsum '
                   'blah blah blah')
```
since strings are automatically continued on newlines if they are within parentheses.

Anyway, thanks! Looks good.

Issue #288, opened 2015-10-26
Send recovery alerts


jasonrhaas:
It is common with server monitoring tools to send a "resolve" message when the problem that triggered the alert has recovered.  It would be nice if there was something in ElastAlert that would send a message out if the next query did not yield the same alert.

For example, if I get a "flatline" alert, I fix the problem, and ElastAlert no longer alerts on that issue, it should send a "recovery" message out to tell whomever is listening that the issue is resolved.

More specifically, I'm using the PagerDuty API to track ElastAlert alerts, and would like to make use of the "resolve events" API.

https://developer.pagerduty.com/documentation/integration/events/resolve

Minjung:
Hi, i talked with the guys on irc about something similar i wanted to implement, basically extending all the alerts that occur over a period of time (frequency, flatline, spike, etc) with something like the EventWindow, but maybe an AlertWindow which starts on the first alert and tracks it until the alert "expires" (ie, no more alerts after x time).
I don't know if i'm going to be able to work on this anytime soon but i just wanted to comment and give a :+1:

bolshoy:
@jasonrhaas I have a simple implementation for this, running it for two weeks already and it looks fine.
The logic is simple: if there is no alert for the rule in the current run, check in the writeback index if we have an alert stored for the previous run of this rule. If there was an alert, try to resolve it. Resolver is implemented by the corresponding alerter. See https://github.com/rounds/elastalert/commit/efc449295636bddee913f4bd3c61d3a857e1d339

jasonrhaas:
thanks @bolshoy for sharing.  Something like this would be a really nice addition to ElastAlert.  It could be another YAML option that is set per alert, like `resolve_alert: true` or something to that effect.

bolshoy:
@jasonrhaas right, this switch might be needed. In our case, alerts are sent to Sensu, and it always needs resolving. Hopefully I'll have some spare time soon l try to create a proper PR.

fiunchinho:
Any news on this?

tomwganem:
Would really like this to be implemented.

eravion:
+1 :)

iainlbc:
+1

Mormaii:
+1

bean5:
+1

bean5:
For flatline, I agree that this is entirely possible, and is something I was looking to do. I would foresee, though, that for other rule types, doing this become more complex, specifically when `query_key` is set.

[...]

PR #1304, opened 2017-08-24
Add support for each PagerDuty event_type (only trigger was supported…

bean5:
Add support for each PagerDuty event_type (only trigger was supported); now supporting:

	* trigger
	* resolve
	* acknowledge


bean5:
You'd want to be careful making alerts than can resolve others, but at least it is supported. This is in response to the idea at #288 by @pblasquez.

bean5:
**Disclaimer**
I just remembered that I wrote a test case for the code, but I did not actually use test this feature against PagerDuty. The PagerDuty API seemed to indicate that both their API versions should accept it, but this remains untested.

PR #1305, opened 2017-08-25
Implement pulling rules from s3

oli77za:
Polls S3 buckets for rules too


oli77za:
Sent the PR too early

PR #1318, opened 2017-09-05
Added resources to build Docker image.

krizsan:
With hope that this shall become the official Elastalert Docker image and make my Elastalert Docker image obsolete.

### DIFF
--- a/Dockerimage/Dockerfile
+++ b/Dockerimage/Dockerfile
@@ -1,5 +1,4 @@
 # Elastalert Docker image running on Alpine Linux.
-# Build image with: docker build -t ivankrizsan/elastalert:latest .
 
 FROM alpine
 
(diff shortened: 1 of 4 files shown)


bean5:
W00t!

Dockerimage/Dockerfile
```diff
@@ -0,0 +1,81 @@
+# Elastalert Docker image running on Alpine Linux.
+# Build image with: docker build -t ivankrizsan/elastalert:latest .
```
bean5: @krizsan you referenced the old repo here.

krizsan: Sorry, missed that one. Fixed.

muravitskiy:
Usually we don't need supervisord when running single app in container. 
Also you should create container pointing to specific release version, not master branch

PR #1329, opened 2017-09-11
fix top_count_keys when using flatline ruletype

aidan-:
Due to the nature of the flatline ruletype, `top_count_keys` always has no results when the rule is triggered.  This is because the code that populates the `top_count_keys` field is only looking during the period that there were no events.  This pull request doubles search window, increasing the likelihood of finding related events.

### DIFF
--- a/elastalert/elastalert.py
+++ b/elastalert/elastalert.py
@@ -1265,7 +1265,16 @@ class ElastAlerter():
                     qk = match[rule['query_key']]
                 else:
                     qk = None
-                start = ts_to_dt(lookup_es_key(match, rule['timestamp_field'])) - rule.get('timeframe', datetime.timedelta(minutes=10))
+
+                if isinstance(rule['type'], FlatlineRule):
+                    # flatline rule triggers when there have been no events from now()-timeframe to now(),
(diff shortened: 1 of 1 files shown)


Qmando:
Sorry for the slow reply here. I imagine this might be confusing for people. What's a scenario this would be useful for?

The alternative is to prevent people from using top_count_keys with flatline at all.

aidan-:
We are currently using the `flatline` rule type to determine if a monitored instance has stopped sending information to Elasticsearch.  This alert is passed through some `match_enhancements` to check to see if the instance is still running (this indicates an issue with the host) or not.

The `top_count_keys` is used to retrieve a couple of additional fields required to be able to query the provider to get the instance status.  An example configuration looks like this:

```
index: metricbeat-*
type: flatline

query_key: meta.cloud.instance_id
doc_type: metricsets

threshold: 1
timeframe:
  minutes: 5

top_count_keys:
  - "meta.cloud.region"
  - "meta.aws.accountid"

raw_count_keys: false

match_enhancements:
  - "ec2_enhancements.EC2SupressIfNotRunning"
```

Some unnecessary lines have been removed from the example configuration.

Qmando:
works for me!

PR #1298, opened 2017-08-22
Thread based rule execution and scheduling

Qmando:
Work in progress!

First pass at adding

1. Threading execution with run_every defined per rule, partially copied from #1128
2. A rule scheduler, inspired by #1114

Note that I haven't *added* any tests yet for these features, I merely fixed the existing tests and have done some manual testing. I mostly wanted to get something up for comment before I continue testing all the edge cases.

PR #1332, opened 2017-09-11
Run time

ruria:
From time to time, we need to run rules only on range time basis, example in working hours. This pull request adds run_time (and run_time_format) to achieve this.

example configuration: running rule all days from 00:00:00 to 01:00:00 and mondays + fridays from 13:00:00 to 20:00:00

run_time:
  - start: "00:00:00"
    end: "01:00:00"
  - start: "13:00:00"
    end: "20:00:00"
    week_day: ["mon", "fri"]


Qmando:
This functionality is also covered by https://github.com/Yelp/elastalert/pull/1298.

PR #1339, opened 2017-09-15
Description field added to elastalaert events

### LINKED ISSUE #662: Add rule_id, reference, and description fields to builtin alerter types
Requesting the addition of three fields.

rule_id: (string, optional, default empty) Arbitrary rule ID defined by the rule author.  Most common rule frameworks have the concept of a rule ID and leverage some kind of numeric or alphanumeric range. 

reference: (string or list of strings, optional, default empty) Arbitrary external reference link(s). Common use case here is one or more hyperlinks to external rule documentation, or to provide outside context pertinent to the rule.

description: this field already exists, but is not available to built-in alerter types.  It should be.

Each of these three fields should be present in the elastalert events, as well as available for inclusion in the built-in alerter types.

ruria:
Partially fix #662

### DIFF
--- a/elastalert/create_index.py
+++ b/elastalert/create_index.py
@@ -10,10 +10,10 @@ import time
 import elasticsearch.helpers
 import yaml
 from auth import Auth
-from envparse import Env
 from elasticsearch import RequestsHttpConnection
 from elasticsearch.client import Elasticsearch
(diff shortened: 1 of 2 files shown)


nsano-rururu:

PR #1351, opened 2017-09-22
fix for must_not

dkamen:
You could not say:
filter:
     must_not: 
              other_criteria

because everything would go under a global "must": array. This fix addresses the issue, essentially exposing the must_not part of the DSL to ElastAlert. Any other type of query is treated as before, going inside the "must" array.


YoavBZ:
You can simply put your `must_not` in a `bool` field instead:
```
filter:
- bool:
  must_not:
  - other_criteria
```

Qmando:
You seem to have removed half of the file as part of the changes.

nsano-rururu:
Test are failing

nsano-rururu:

PR #1386, opened 2017-10-17
feature run elastalert in new relic style

NikolaMandic:
this has no tests
and no docs

but it uses file system locks to provide feature of having alert
alerting on start of issue and on stop of the issue
in similar fashion as new relic

if you put
 
```
alert_once: true
alert_lock_key: hostname
```

it will trigger alerts once per hostname when alert is first seen and when alert is over


nsano-rururu:
Test are failing

nsano-rururu:

PR #1397, opened 2017-10-24
Clear jira args on calling get_aribitrary_fields

Qmando:
This allows for an enhancement to set and clear arbitrary jira fields. Previously, if get_arbitrary_fields was called, it would add to jira_args but never remove any. Now, it clear jira_args which allows for changes. Note than many fields are still stuck because they are set in __init__ and cannot be changed.

### DIFF
--- a/elastalert/alerts.py
+++ b/elastalert/alerts.py
@@ -548,7 +548,7 @@ class JiraAlerter(Alerter):
             self.get_arbitrary_fields()
         except JIRAError as e:
             # JIRAError may contain HTML, pass along only first 1024 chars
-            raise EAException("Error connecting to JIRA: %s" % (str(e)[:1024]))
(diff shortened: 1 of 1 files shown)


elastalert/alerts.py
```diff
@@ -539,6 +540,26 @@ def __init__(self, rule):
...
 
+        self.reset_jira_args()
+
+        try:
+            self.client = JIRA(self.server, basic_auth=(self.user, self.password))
+            self.get_priorities()
+            self.get_arbitrary_fields()
+        except JIRAError as e:
+            # JIRAError may contain HTML, pass along only first 1024 chars
+            raise EAException("Error connecting to JIRA: %s" % (str(e)[:1024]))
```
danielpops: Don't do this, it loses the inner stack. Use the overload of raise that takes the `sys.exc_info()[2]`

danielpops:
Just curious, what bug does this address? Is it a threading issue or is this something that's been broken for a long time?

Issue #996, opened 2017-04-07
Another failed to parse date field error


joaoantoniopereira:
Hi,
I saw many error like mine but somehow I cannot overcome this:

Config file:
```
es_host: localhost
es_port: 9200
type: spike
index: metrics
threshold_cur: 1
threshold_ref: 1
spike_type: "both"
filter:
- query:
    query_string:
      query: "M_kernel:>0"
alert:
- "email"
email:
- "my email"
```
While running elastalert-test-rule i get:

##############################################################3
Successfully loaded Event spike

INFO:requests.packages.urllib3.connectionpool:Starting new HTTP connection (1): localhost
Got 2 hits from the last 1 day

Available terms in first hit:
	M_led_azul_dist
	M_upgrades_downloadf_dist
	M_led_roxa
	@timestamp
	M_kernel

INFO:elastalert:Note: In debug mode, alerts will be logged to console but NOT actually sent. To send them, use --verbose.
INFO:requests.packages.urllib3.connectionpool:Starting new HTTP connection (1): localhost
INFO:requests.packages.urllib3.connectionpool:Starting new HTTP connection (1): localhost
INFO:elastalert:Queried rule Event spike from 2017-04-06 11:18 WEST to 2017-04-06 11:33 WEST: 0 / 0 hits
INFO:elastalert:Queried rule Event spike from 2017-04-06 11:33 WEST to 2017-04-06 11:48 WEST: 0 / 0 hits
INFO:elastalert:Queried rule Event spike from 2017-04-06 11:48 WEST to 2017-04-06 12:03 WEST: 0 / 0 hits
INFO:elastalert:Queried rule Event spike from 2017-04-06 12:03 WEST to 2017-04-06 12:18 WEST: 0 / 0 hits
INFO:elastalert:Queried rule Event spike from 2017-04-06 12:18 WEST to 2017-04-06 12:33 WEST: 0 / 0 hits
INFO:elastalert:Queried rule Event spike from 2017-04-06 12:33 WEST to 2017-04-06 12:48 WEST: 0 / 0 hits
INFO:elastalert:Queried rule Event spike from 2017-04-06 12:48 WEST to 2017-04-06 13:03 WEST: 0 / 0 hits
INFO:elastalert:Queried rule Event spike from 2017-04-06 13:03 WEST to 2017-04-06 13:18 WEST: 0 / 0 hits
[...]

PR #1405, opened 2017-10-27
Fix dt_to_unix to return a int and avoid scientific notation on serialization

brunotm:
This change fixes the epoch_seconds searialization to avoid scientific
notations and parse errors with Elasticsearch as seen before on #996.

### DIFF
--- a/elastalert/util.py
+++ b/elastalert/util.py
@@ -233,7 +233,7 @@ def unix_to_dt(ts):
 
 
 def dt_to_unix(dt):
-    return total_seconds(dt - datetime.datetime(1970, 1, 1, tzinfo=dateutil.tz.tzutc()))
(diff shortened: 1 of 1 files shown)


Qmando:
Aha! I wasn't able to reproduce this with seconds OR millisecond epochs, now I see if I make it large enough, I can trigger it. Do you have any idea if this is system dependent? 
```
>>> json.dumps({'hello': 14914991344305444.1})
'{"hello": 1.4914991344305444e+16}'
>>> json.dumps({'hello': 1491499134430544.1})
'{"hello": 1491499134430544.0}'
```

brunotm:
@Qmando 
>Do you have any idea if this is system dependent?

Don't really know, but I hit this in all deployments we have.

Thanks for merging.

PR #1417, opened 2017-11-09
Stomp alert Update

snuids:
Stomp alerting code changed in order to correctly display the query_key matches when the query key is complex. (The query_key is a sub property of a property such as: alert.type)

### DIFF
--- a/elastalert/alerts.py
+++ b/elastalert/alerts.py
@@ -294,17 +294,18 @@ class Alerter(object):
 
 class StompAlerter(Alerter):
     """ The stomp alerter publishes alerts via stomp to a broker. """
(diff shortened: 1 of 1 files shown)


elastalert/alerts.py
```diff
@@ -296,20 +296,30 @@ class StompAlerter(Alerter):
     """ The stomp alerter publishes alerts via stomp to a broker. """
     required_options = frozenset(['stomp_hostname', 'stomp_hostport', 'stomp_login', 'stomp_password'])
 
-    def alert(self, matches):
+    def check_if_match_exists(mes,matchkey):
```
Qmando: This function is a duplicate of util.lookup_es_key.

snuids: Right, I will update my code to use util.lookup_es_key. Thx for the hint.

Issue #1221, opened 2017-07-12
is_enabled for rules?


yuanxu-li:
Here is the problem. I have defined many rules, but sometimes I want to disable some of them according to needs. 

The current way is to delete not used ones from rules folder and add them back when needed, and simply run:
`python -m elastalert.elastalert --verbose`
which is a bit silly.

I know I can also write another script to solve it, but I just wonder if there is any attribute that we can define in a rule to easily enable or disable this rule.

Thanks.

Qmando:
You can rename them and add ".disabled" or something to the end of the file name. There's nothing you can add in the rule itself to disable it.

jimbocoder:
I have been using exactly that approach, renaming to `.disabled` but it's awkward in my opinion. A toggle with something like `is_enabled` that would live-reload would be a very welcome addition.

Qmando:
I'd accept a pull request for this feature. Just add some kind of short circuit in config.py.

jimbocoder:
Cool, thanks. Drawing @jasonhkennedy's attention here as well in case he's got time and interest here before I can work on this.

PR #1428, opened 2017-11-16
add is_enabled to rule attribute for a rule to be easily disabled

yul-yelp:
fix https://github.com/Yelp/elastalert/issues/1221

Let `is_enabled` be an attribute of rule, and disable a rule easily by setting `is_enabled: False` in a rule file.

Filter rules in three places:
1. when elastalert client starts, and it calls load_rules to initialize rules
2. when elastalert syncs rules, and it calls load_rules_changes to check current rules
3. when elastalert syncs rules, and it calls load_rules_changes to load newly added rules

### DIFF
--- a/elastalert/config.py
+++ b/elastalert/config.py
@@ -462,6 +462,9 @@ def load_rules(args):
     for rule_file in rule_files:
         try:
             rule = load_configuration(rule_file, conf, args)
+            # By setting "is_enabled: False" in rule file, a rule is easily disabled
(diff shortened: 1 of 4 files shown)


gzeronet:
:+1:

jkone27:
is there examples in docs for this? 

i don't see it in the Yelp schema def:
https://github.com/Yelp/elastalert/blob/master/elastalert/schema.yaml

PR #1443, opened 2017-12-01
Treat summary_table cell values as text

danielpops:
This makes it so that the \_\_str\_\_() of each field is populated in the cells. Without this change, Texttable tries to figure out "automatically" what the type is and represent it the way it thinks is "best". In some cases, it's not the best (e.g. for an AccountId).

I suppose this behavior could be made configurable. I'll leave that exercise to the next person that wants to change the behavior :)

I could also see the argument for NOT making the `count` column text, and to make it automatic, since in that case, a very large number might make sense in scientific notation, but ::shrug:: for now, unless reviewers feel strongly about it.

How I know this works:

```
>>> from texttable import Texttable
>>> t1 = Texttable()
>>> t1.header(['AccountId'])
>>> t1.add_row([1234567890])
>>> print t1.draw()
+----------------+
|    AccountId   |
+================+
| 1.235e+09      |
+----------------+
>>> t2 = Texttable()
>>> t2.header(['AccountId'])
>>> t2.set_cols_dtype(['t'])
>>> t2.add_row([1234567890])
>>> print t2.draw()
+----------------+
|    AccountId   |
+================+
| 1234567890     |
+----------------+
```

### DIFF
--- a/elastalert/alerts.py
+++ b/elastalert/alerts.py
@@ -262,6 +262,8 @@ class Alerter(object):
             )
             text_table = Texttable(max_width=self.get_aggregation_summary_text__maximum_width())
             text_table.header(summary_table_fields_with_count)
+            # Format all fields as 'text' to avoid long numbers being shown as scientific notation
+            text_table.set_cols_dtype(['t' for i in summary_table_fields_with_count])
             match_aggregation = {}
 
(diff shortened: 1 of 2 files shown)


danielpops:
BTW, that test failure doesn't look to be related to my changes, though I can repro locally.

PR #1461, opened 2017-12-14
hipchat: support @mention specific users on Hipchat

multani:
When sending HTML messages to Hipchat, we can't use @mention to ping users, due to limitation (feature) in the Hipchat API.

This new option, if defined in the configuration, will first pings the users specified using a "text" message format (which triggers the ``@mentions`` mechanism and notify users directly, if configured as such) then send the actual HTML message.


Qmando:
Looks good. Can you add a section in the documentation?

multani:
> Can you add a section in the documentation?

Sure, done!

PR #1472, opened 2018-01-01
Compatibility with Elastic 6

snuids:
Hi,

I followed the same idea as Jensvandecasteele and duplicated the indexes in order to avoid the doc type problem. I simply added a few tests in order to keep the compatibility with Elastic versions 4 and 5. The multi indexes are only used if ELK>=6

Cheers,


changi67:
Patch is working for me on ES6.0.0

CodingSpiderFox:
I hope this will get merged soon ;)

PR #1476, opened 2018-01-05
Fixed a bug causing a TypeError when nested fields were null

Qmando:
if dict_cursor became None and there were more subkeys, a TypeError would be thrown.


ketanyelp:
ship it!

PR #1485, opened 2018-01-11
fix #1481 add missing syntax for Kibana 3

### LINKED ISSUE #1481: filter download_dashboard AttributeError: 'dict' object has no attribute 'insert'
using elastalert 0.1.26 and kibana 3 I get an error trying to download filters from kibana 3:
`es_filters['filter']['bool']['must'].insert(0, {'range': {timestamp_field: {'gt': starttime,
AttributeError: 'dict' object has no attribute 'insert`
I am using command `elastalert-test-rule --config config.yaml rules/006-Dashboard.yaml `
I get message `Successfully loaded Dashboard` but after that I get an error above
My rule file has these settings:
`type: frequency`
`num_events: 1`
`timeframe:
  minutes: 5`
`filter:
  download_dashboard: "dashboardname"`

andrpetr:


Qmando:
Hm my comment that issue was actually completely wrong. That feature was just broken. Sorry about that.

You can still use ``elastalert-rule-from-kibana`` to generate the filter and then copy it over.

PR #1486, opened 2018-01-11
Query key value length option for FrequencyRule

krrrr38:
## Motivation

Sometimes messages have different value fields, such as followings.

```
{"unique_key": "this is error message: value=1"}
{"unique_key": "this is error message: value=2"}
{"unique_key": "this is error message: value=3"}
```

I want to count frequency by prefix value `this is error message` as a unique key. 

## How to change

Add `query_key_value_length` option into FrequencyRule and omit value by the length.

### DIFF
--- a/docs/source/ruletypes.rst
+++ b/docs/source/ruletypes.rst
@@ -118,6 +118,8 @@ Rule Configuration Cheat Sheet
 +----------------------------------------------------+--------+-----------+-----------+--------+-----------+-------+----------+--------+-----------+
 | ``query_key`` (string, no default)                 |   Opt  |           |           |   Req  |    Opt    |  Opt  |   Opt    |  Req   |  Opt      |
 +----------------------------------------------------+--------+-----------+-----------+--------+-----------+-------+----------+--------+-----------+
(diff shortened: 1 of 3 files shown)


Qmando:
I'm not totally opposed to this idea, though generally I'd say transformations like this should be done before the document is ingested into logstash, not by elastalert.

Also, your code only applies if you're using frequency type and ``use_terms_query``, but there's no reason it shouldn't work for other rule types and query types.

krrrr38:
Thank you for a quick response.

> this should be done before the document is ingested into logstash, not by elastalert

It's possible to throw such kind of additional data. But in our case, many application enqueue data into ElasticSearch, so it's difficult to change all applications.
And I think this kind of alert aggregation features are responsible for elastalert layer.

> there's no reason it shouldn't work for other rule types and query types.

Sorry, it's my laziness. If you can agree my suggestion, I will try to apply this to other rule types.
thanks.

krrrr38:
I'm not familier with `use_terms_query` logic now, so please let me give a time to check them.
When I understand them, I will open pr or something again. close this pr temporally.

PR #1487, opened 2018-01-11
Stylize alert_text using HTML in StrideAlerter 

jness:
Many of my existing elastalert rules take advantage of HipChat's ability to stylize html correctly in the client. However, Stride uses a specialized document format, and doesn't have the ability to process html:

https://developer.atlassian.com/cloud/stride/apis/document/structure/

Below is an example of alert_text and formatting common found in many of my rules.

```
alert_text: "<b>{0}</b>: At least {1} errors occurred for program_type {2}. <a href='{3}'>Kibana Link</a>"
alert_text_type: alert_text_only

alert_text_args:
  - name
  - num_events
  - program_type
  - kibana_link
```

This pull request uses python's standard libraries HTMLParser, then converts html tags into Stride document structure.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,20 @@
 # Change Log
 
+## v0.1.26
+
+### Added
+- Added support for Elasticsearch 6
+- Added support for mentions in Hipchat
+
+### Fixed
+- Fixed an issue where a nested field lookup would crash if one of the intermediate fields was null
+
+## v0.1.25
+
+### Fixed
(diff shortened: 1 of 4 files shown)


kakovskyi:
I think that the patch makes ElastAlert is actually usable with Stride, not only proof of concept as it was before. Now we can put links to Kibana to see actual events instead of just knowing that _Something went wrong_.

We applied the patch and now use it for reals. Would like to share that with the community.
Thank you, @jness, for making that possible.

PR #1493, opened 2018-01-14
add slack_disable_escape_body option

krrrr38:
## Problem

Generally [Slack webhook accept hypertext format](https://api.slack.com/docs/message-formatting#linking_to_urls) like `<http://www.foo.com|www.foo.com>`. But elastalert automatically escape `<` and `>`. So, if we send message with `<http://www.foo.com|www.foo.com>`, not become links like followings.

https://github.com/Yelp/elastalert/blob/46d807b3720f3046c3290b019ec0edb2b4f0846a/elastalert/alerts.py#L1049-L1055

- `slack_override_parse: none`

image

- `slack_override_parse: full`

image

## How to change?

add `slack_disable_format_body` option, which disable elastalert formatting.

image

### DIFF
--- a/docs/source/ruletypes.rst
+++ b/docs/source/ruletypes.rst
@@ -1503,7 +1503,7 @@ Provide absolute address of the pciture, for example: http://some.address.com/im
 
 ``slack_proxy``: By default ElastAlert will not use a network proxy to send notifications to Slack. Set this option using ``hostname:port`` if you need to use a proxy.
 
-``slack_disable_format_body``: By default ElastAlert will format slack body automatically. Set this option ``True`` if you want to disable formatting.
+``slack_disable_escape_body``: By default ElastAlert will escape slack body automatically. Set this option ``True`` if you want to disable escape.
 
 Telegram
 ~~~~~~~~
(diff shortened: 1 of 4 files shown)


jrauschenbusch:
I've two questions to you PR:
- As the Slack API docs states all messages must be UTF-8 encoded: Won't this also be disabled with your code changes?
- In which case a general escaping of the special characters makes really sense? I think the goal should be to never escape any kind of control characters (https://api.slack.com/docs/message-formatting#control_sequences). If one wants to really render the characters, he/she needs to use the escaped character.

krrrr38:
> As the Slack API docs states all messages must be UTF-8 encoded: Won't this also be disabled with your code changes?

UTF-8 encoded is required. so I changed the name to `slack_disable_escape_body`.

> In which case a general escaping of the special characters makes really sense? I think the goal should be to never escape any kind of control characters (https://api.slack.com/docs/message-formatting#control_sequences). If one wants to really render the characters, he/she needs to use the escaped character.

I agree that not escaping body should be a default. In this pr, currently I just keep old behavior.

krrrr38:
ah, you worked at #1497 and it looks good to me. if it merged, I don't need this pr now.

krrrr38:
#1497 was merged, so close this pr.

Issue #660, opened 2016-08-01
Slack and Kibana_link


remotesyssupport:
We predominantly use slack as our alert notifications. I wanted to know if it is possible to include the kibana4 dashboard link in the slack alerts 

My check with including it in alert_text_args (as shown below) doesnt seem to work. Although if I use alert_text_type as default, I can see the kibana_link in the fields of the slack notification

```
# Kibana Link
use_kibana4_dashboard: "https://kibana.internal/app/kibana#/dashboard/applog"
kibana4_start_timedelta:
  minutes: 10
kibana4_end_timedelta:
  minutes: 10

# Alert Config
alert:
- "slack"

alert_text_type: exclude_fields
alert_text_args:
- kibana_link
```

Qmando:
You have to set ``alert_text``.

``alert_text: {0} is the kibana link``

Qmando:
Try ``<{0}|Link to Kibana>`` from https://twitter.com/slackapi/status/579771083181699073

remotesyssupport:
Nope : (
It does exactly as is stated in the latest reply to the comment by Malone Hedges

remotesyssupport:
@Qmando 
It shows up like this
![image](https://cloud.githubusercontent.com/assets/705870/17306884/f0fb9a20-584f-11e6-93ee-a230867684c2.png)

remotesyssupport:
@Qmando 
For now (till this issue is found) got a duck-tape solution, as we want the nice formatted message sent <At least xxx event occurred ...> along with the links. Combined the slack and command alerts. Command sends the kibana link using Slacktee.

```
# Alert Config
alert:
- "slack"
- "command"

alert_text_type: exclude_fields

command: ["/opt/elastalert/bin/send_via_slacktee", "<%(kibana_link)s| App Errors Kibana Link>"]
```

Output:

![image](https://cloud.githubusercontent.com/assets/705870/17308146/1f908930-5856-11e6-8118-a2fef88427b9.png)

Question: Is Number of events detected, StartTime and EndTime available as variable like kibana_link? If yes, we can then only use Command Alert only.

schickling:
@Qmando could you please reopen this issue as the formatting doesn't work as expected?

schickling:
Thanks @Qmando. Is there an easy fix for this? Would be super awesome if this would work. :shipit:

brdaugherty:
Including a clickable URL to a kibana dashboard is working for me in v0.0.95.

[...]

PR #1497, opened 2018-01-15
Fix wrong link encoding in slack message body

jrauschenbusch:
Removes escaping of `>`, `<` and `&` characters from slack message body as this leads to non-interpreted links as described here: https://github.com/Yelp/elastalert/issues/660#issuecomment-315151378

### DIFF
--- a/elastalert/alerts.py
+++ b/elastalert/alerts.py
@@ -1048,11 +1048,7 @@ class SlackAlerter(Alerter):
 
     def format_body(self, body):
         # https://api.slack.com/docs/formatting
-        body = body.encode('UTF-8')
-        body = body.replace('&', '&amp;')
-        body = body.replace('<', '&lt;')
-        body = body.replace('>', '&gt;')
-        return body
+        return body.encode('UTF-8')
 
     def get_aggregation_summary_text__maximum_width(self):
         width = super(SlackAlerter, self).get_aggregation_summary_text__maximum_width()


Qmando:
From that comment, it appears that this only fixes the behavior when parse=full. Maybe we should also change the default for slack_parse_override from 'none' to 'full' ?

jrauschenbusch:
I think the default parsing mode should still be `parse=none` or `parse=` (empty) field. Look at this examples:
https://api.slack.com/docs/message-formatting#parsing_modes

The examples provided there explain it really good:

1. If no `parse` mode is used, then all special control sequences like `<#channel>` are interpreted, but also plain URLs like `https://github.com`. (slack default behavior)
2. If `parse=none` is used, then special control sequences like `<#channel>` and links in form of `<https://github.com>` are interpreted, but plain URLs like `https://github.com` aren't.
3. In `parse=full` is used, all special control sequences are ignored and maybe in the output HTML-encoded.

If you are use mode 1. or 2 and you don't want to render special control sequences (e.g. like <#channel> or <@person>, you have to HTML-encode the control chars of the specific message part:
`<http://github.com>` must then be encoded to `&lt;http://github.com&gt;`

If you are using mode 3 (`parse=full`), all special control sequences will be ignored and are HTML-encoded by slack.

You can also play around with this tool which uses the default slack behaviour (i.e. no `parse` mode):
https://api.slack.com/docs/messages/builder?msg=%7B%22text%22%3A%20%22This%20%26amp%3B%20that.%20%26lt%3B%20and%20%26gt%3B%22%7D

In this case,

Qmando:
I see, that makes sense. It's not currently possible to NOT send a parse value, but most uses cases will still be possible after this PR.

krrrr38:
@Qmando could you release new version?

Qmando:
Sure

Issue #1480, opened 2018-01-08
Problem when restarting elastalert with change rule


Joao-Andrade:
Hello

I'm having a problem when, for example, the elastalert instance goes down and after a while, it starts again. If i have a ChangeRule type, the events that occurred during the time that elastalert was down will not be considered and there will be no alerts regarding those. And, for example, if a key had a certain value during that time, and when the elastalert starts again, the value is changed, it will not alert because it has no history backwards of the values.

One solution i found was adding old_query_limit field as 0 days. But that raises the problem of having duplicate  matches.

Is it possible to check backwards in time, for example, until timeframe and have an history of the occurrences of the keys, so when elastalert runs the rule for the first time, it has that information and when checks for changes, it alerts if the value was changed?
I think this could be useful for other rule types as well. For example, SpykeRule where it could know if there was a spyke during that period.

Thank you
João Andrade

PR #1499, opened 2018-01-16
Fix to ChangeRule to allow elastalert to search backwards for the last occurrences

Joao-Andrade:
If i have a ChangeRule type, and elastalert goes down for some time, the events that occurred during the time that elastalert was down will not be considered and there will be no alerts regarding those. And, for example, if a key had a certain value during that time, when elastalert starts again, the value is changed and it will not alert because it has no history of the values for that key.
Now, if the rule has the parameter "check_last_occurrences" and elastalert is started or the rule is modified, elastalert will fill in the occurrences variable with the last values from “timeframe” ago until now, so the next time a document with the query key appears, and the compare key is changed, an alert will be sent.

This also fixes issue #1480 .

### DIFF
--- a/elastalert/elastalert.py
+++ b/elastalert/elastalert.py
@@ -861,7 +861,7 @@ class ElastAlerter():
         self.writeback('elastalert_status', body)
 
         return num_matches
-		
+
     def init_change_rule_occurrences(self, rule):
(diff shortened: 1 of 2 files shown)


brunotm:
➕ 1

nsano-rururu:

PR #1503, opened 2018-01-19
Stomp Alert Fixes and Improvements

snuids:
* Better JSON format of the Stomp Alert. 
* Add a 1 second delay in order to receive the CONNECTED frame back from the server before disconnecting.  It avoids a possible connection leak with the broker. In a perfect world, this one should be fixed in the stomp.py library.


Qmando:
This probably needs a little bit more control of what exactly goes into the JSON body of the alert, but looks good for now.

PR #1544, opened 2018-02-09
Index alert text

tomrade:
Added Alert Text to the Elasticsearch Index, which is useful for teams / tools which leverage the index as their alert database, and for looking back over old alerts


tomrade:
this can get awkward with alert bodys that add the fields

PR #1545, opened 2018-02-09
Add Alert_Text to ES index, reused existing method from alerts.py

tomrade:
Reused the method from Alerts.py rather than reimplement the same thing again , however I used a cloned Rule object with the 'alert_text_only' option set as otherwise the match fields would end up here


nsano-rururu:

Issue #1053, opened 2017-04-25
Support for fields in Slack alerts


buinauskas:
Slack incoming webhooks support an array of fields. [Example](https://api.slack.com/docs/messages/builder?msg=%7B%22attachments%22%3A%5B%7B%22fallback%22%3A%22Required%20plain-text%20summary%20of%20the%20attachment.%22%2C%22color%22%3A%22%2336a64f%22%2C%22pretext%22%3A%22Optional%20text%20that%20appears%20above%20the%20attachment%20block%22%2C%22author_name%22%3A%22Bobby%20Tables%22%2C%22author_link%22%3A%22http%3A%2F%2Fflickr.com%2Fbobby%2F%22%2C%22author_icon%22%3A%22http%3A%2F%2Fflickr.com%2Ficons%2Fbobby.jpg%22%2C%22title%22%3A%22Slack%20API%20Documentation%22%2C%22title_link%22%3A%22https%3A%2F%2Fapi.slack.com%2F%22%2C%22text%22%3A%22Optional%20text%20that%20appears%20within%20the%20attachment%22%2C%22fields%22%3A%5B%7B%22title%22%3A%22Priority%22%2C%22value%22%3A%22High%22%2C%22short%22%3Atrue%7D%2C%7B%22title%22%3A%22Priority%22%2C%22value%22%3A%22High%22%2C%22short%22%3Atrue%7D%2C%7B%22title%22%3A%22Priority%22%2C%22value%22%3A%22High%22%2C%22short%22%3Atrue%7D%5D%2C%22image_url%22%3A%22http%3A%2F%2Fmy-website.com%2Fpath%2Fto%2Fimage.jpg%22%2C%22thumb_url%22%3A%22http%3A%2F%2Fexample.com%2Fpath%2Fto%2Fthumb.png%22%2C%22footer%22%3A%22Slack%20API%22%2C%22footer_icon%22%3A%22https%3A%2F%2Fplatform.slack-edge.com%2Fimg%2Fdefault_application_icon.png%22%2C%22ts%22%3A123456789%7D%5D%7D)

```json
{
  "attachments": [
    {
      "fallback": "Required plain-text summary of the attachment.",
      "color": "#36a64f",
      "pretext": "Optional text that appears above the attachment block",
      "author_name": "Bobby Tables",
      "author_link": "http://flickr.com/bobby/",
      "author_icon": "http://flickr.com/icons/bobby.jpg",
      "title": "Slack API Documentation",
[...]

PR #1552, opened 2018-02-12
Adding Slack fields support

adamenger:
This addresses https://github.com/Yelp/elastalert/issues/1053 and makes incoming alerts a bit easier on the eyes. Please let me know if I can improve this patch at all. The `title` is the name of the Field and then the value is the elasticsearch key you'd like to populate as the value.

Given this config:
```
slack_alert_fields:
- title: First Field
  value: columns.address
  short: true
- title: Second Field
  value: decorations.username
  short: true
```

and this patch, here is the output:

test

Cheers,

### DIFF
--- a/elastalert/alerts.py
+++ b/elastalert/alerts.py
@@ -1060,6 +1060,7 @@ class SlackAlerter(Alerter):
         self.slack_msg_color = self.rule.get('slack_msg_color', 'danger')
         self.slack_parse_override = self.rule.get('slack_parse_override', 'none')
         self.slack_text_string = self.rule.get('slack_text_string', '')
+        self.slack_alert_fields = self.rule.get('slack_alert_fields', '')
 
     def format_body(self, body):
         # https://api.slack.com/docs/formatting
@@ -1075,6 +1076,14 @@ class SlackAlerter(Alerter):
         if text:
             text = u'```\n{0}```\n'.format(text)
         return text
+
+    def populate_fields(self, matches):
+        alert_fields = []
+        for arg in self.slack_alert_fields:
+            arg = copy.copy(arg)
+            arg['value'] = lookup_es_key(matches[0], arg['value'])
+            alert_fields.append(arg)
+        return alert_fields
 
     def alert(self, matches):
         body = self.create_alert_body(matches)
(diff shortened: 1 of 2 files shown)


adamenger:
Hi @jotterbot 

I have noticed the same issue - any help you can give regarding this issue would be really helpful. This was a quick patch on my end to add support but I'm not an expert python developer by any means :) Thanks for giving it a try!

elastalert/alerts.py
```diff
@@ -1076,6 +1077,15 @@ def get_aggregation_summary_text(self, matches):
             text = u'```\n{0}```\n'.format(text)
         return text
 
+    def populate_fields(self, matches):
+        slack_alert_fields = self.rule['slack_alert_fields']
+        alert_fields = []
+        if 'slack_alert_fields' in self.rule:
```
Qmando: This is unnecessary.

elastalert/alerts.py
```diff
@@ -1076,6 +1077,15 @@ def get_aggregation_summary_text(self, matches):
             text = u'```\n{0}```\n'.format(text)
         return text
 
+    def populate_fields(self, matches):
+        slack_alert_fields = self.rule['slack_alert_fields']
```
Qmando: This is unnecessary

elastalert/alerts.py
```diff
@@ -1076,6 +1077,15 @@ def get_aggregation_summary_text(self, matches):
...
+        alert_fields = []
+        if 'slack_alert_fields' in self.rule:
+            for arg in slack_alert_fields:
```
Qmando: ``for arg in self.slack_alert_fields``

elastalert/alerts.py
```diff
@@ -1076,6 +1077,15 @@ def get_aggregation_summary_text(self, matches):
...
+                arg['value'] = lookup_es_key(matches[0], arg['value'])
```
Qmando: Add this before line 1085 to fix the issue reporter about subsequent alerts failing.
```
arg = copy.copy(arg)
```

Qmando:
So sorry about no response to this for so long. 

Once you address these small comments, I'm happy to merge

adamenger:
Thanks @Qmando for your feedback! Can you let me know if this addresses your concerns?

PR #1590, opened 2018-03-06
Fix sending telegram cropped message

enkov:
We need to use unicode string in cropped message.


tjjjwxzq:
Is this going to be merged in any time soon?

PR #1598, opened 2018-03-12
[RFC] writeback_alias added as required global config value

pdscopes:
I've been wanting to separate out my alerts into different indexes, for example by date.
This is working attempt of this but doesn't have the necessary documentation updates.

I've added a new `writeback_alias` as a required value in `required_globals`. An alias to the `elastalert` index (defaults for `elastalert_alerts`) and a template for `elastalert_*` indices that can be generated on the fly. Elastalert then uses the alias when searching for alerts.

Rules have a new optional value of `writeback_suffix`. This can include a timestamp. For example:
`writeback_suffix: %Y.%d` would mean that your alerts would be put into an index like `elastalert_2018.03`.

I would love feedback/comments on this! Where this could be improved. I've tried to ensure it is backwards compatible but I have not tested in ES2 (or ES6)

### DIFF
--- a/elastalert/config.py
+++ b/elastalert/config.py
@@ -29,7 +29,8 @@ from util import unixms_to_dt
 rule_schema = jsonschema.Draft4Validator(yaml.load(open(os.path.join(os.path.dirname(__file__), 'schema.yaml'))))
 
 # Required global (config.yaml) and local (rule.yaml)  configuration options
-required_globals = frozenset(['run_every', 'rules_folder', 'es_host', 'es_port', 'writeback_index', 'buffer_time'])
+required_globals = frozenset(['run_every', 'rules_folder', 'es_host', 'es_port', 'writeback_index',
+                              'writeback_alias', 'buffer_time'])
 required_locals = frozenset(['alert', 'type', 'name', 'index'])
 
 # Settings that can be derived from ENV variables
--- a/elastalert/create_index.py
+++ b/elastalert/create_index.py
@@ -32,6 +32,7 @@ def main():
     parser.add_argument('--verify-certs', action='store_true', default=None, help='Verify TLS certificates')
     parser.add_argument('--no-verify-certs', dest='verify_certs', action='store_false', help='Do not verify TLS certificates')
     parser.add_argument('--index', help='Index name to create')
+    parser.add_argument('--alias', help='Alias name to create')
     parser.add_argument('--old-index', help='Old index name to copy')
     parser.add_argument('--send_get_body_as', default='GET', help='Method for querying Elasticsearch - POST, GET or source')
     parser.add_argument(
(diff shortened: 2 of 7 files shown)


pdscopes:
I've now updated the tests so that they pass. Also I have tested this in ES v6 and it works.

Sedor:
Would be great if this is merged into the project.

Qmando:
Thank you for the patience and excellent work. I will review and merge this ASAP

pdscopes:
There will need to be changes made to the documentation.

Also, I believe indices can be safely created/recreated as the alias can catch the other indices (silence, error, etc) if not removed before indices are created/recreated.

Sedor:
@Qmando  Any progress here ?
Or is there i way i could help ?

elastalert/create_index.py
```diff
@@ -250,19 +258,25 @@ def main():
...
+        es.indices.put_template(name='elastalert', body={'index_patterns': ['elastalert_*'],
+                                                         'aliases': {alias: {}}, 'mappings': es_mapping})
         print('New index %s created' % index)
     else:
         es.indices.put_mapping(index=index, doc_type='elastalert', body=es_mapping)
         es.indices.put_mapping(index=index, doc_type='elastalert_status', body=ess_mapping)
         es.indices.put_mapping(index=index, doc_type='silence', body=silence_mapping)
         es.indices.put_mapping(index=index, doc_type='elastalert_error', body=error_mapping)
         es.indices.put_mapping(index=index, doc_type='past_elastalert', body=past_mapping)
+        es.indices.put_alias(index=index, name=alias, body={'filter': {'term': {'_type': 'elastalert'}}})
```
Qmando: What does adding a filter do an index do?

elastalert/config.py
```diff
@@ -29,7 +29,8 @@
 rule_schema = jsonschema.Draft4Validator(yaml.load(open(os.path.join(os.path.dirname(__file__), 'schema.yaml'))))
 
 # Required global (config.yaml) and local (rule.yaml)  configuration options
-required_globals = frozenset(['run_every', 'rules_folder', 'es_host', 'es_port', 'writeback_index', 'buffer_time'])
+required_globals = frozenset(['run_every', 'rules_folder', 'es_host', 'es_port', 'writeback_index',
+                              'writeback_alias', 'buffer_time'])
```
Qmando: Can this be optional? I'd like to not introduce breaking changes. I think it's OK to create the index alias by default, but i'd also like to see ``writeback_alias`` have a default value, "elastalert_alerts" is probably ok.

pdscopes: Yes of course, it would make a lot more sense to have `writeback_alias` as an optional value.

elastalert/elastalert.py
```diff
@@ -199,19 +200,27 @@ def get_index(rule, starttime=None, endtime=None):
...
+                elif doc_type == 'past_elastalert':
+                    writeback_index += '_past'
+                elif doc_type == 'elastalert_status':
+                    writeback_index += '_status'
+                elif doc_type == 'elastalert_error':
+                    writeback_index += '_error'
+        else:
+            suffix = rule['writeback_suffix']
+            if '%' in rule['writeback_suffix']:
+                format_start = suffix.find('%')
```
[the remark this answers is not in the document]
pdscopes: I'll give it a go

pdscopes:
I was just wondering if this PR is going to be accepted?

PR #1627, opened 2018-04-03
Configure ElasticSearch URL prefix using environment variable

AndreLouisCaron:
I run `elastalert` inside a container against multiple environments and the value I need to put in `es_url_prefix` changes depending on the environment.

Currently, I need to write a new `config.yaml` file for each environment and bind-mount it into the container, which is pretty tedious.  I can already configure `es_host`, `es_port` and `use_ssl` using environment variables, so I though doing the same for `es_url_prefix` would be neat.

Let me know what you think :-)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ build/
 
 my_rules
 *.swp
+*~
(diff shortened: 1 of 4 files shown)


AndreLouisCaron:
@Qmando I don't mean to be pushy (I assume you are very busy with other matters), but I'm still hoping to get this minor change merged as it solves a workflow issue for me.  Any chance you can take a minute to look at this?

Thanks!

PR #1630, opened 2018-04-03
Silence "Expected str timestamp, got datetime" warnings

### LINKED ISSUE #1629: WARNING:root:Expected str timestamp, got datetime
I'm getting these warnings when running `elastalert==0.1.29`:

    WARNING:root:Expected str timestamp, got datetime

I took a quick look at the code and it seems like they come from `elastalert/util.py`:

```python
def ts_to_dt(timestamp):
    if isinstance(timestamp, datetime.datetime):
        logging.warning('Expected str timestamp, got datetime')
        return timestamp
    dt = dateutil.parser.parse(timestamp)
    # Implicitly convert local timestamps to UTC
    if dt.tzinfo is None:
        dt = dt.replace(tzinfo=dateutil.tz.tzutc())
    return dt
```

Is there any reason for this warning?  If needed, I can send in a pull request to silence it :-)

AndreLouisCaron:
The caller is asking for a datetime object and the result is
already a datetime object.  It's not clear why we would want
to warn about this and the warnings are triggered during "normal"
elastalert operation.

Fixes #1629.

### DIFF
--- a/elastalert/util.py
+++ b/elastalert/util.py
@@ -108,7 +108,6 @@ def lookup_es_key(lookup_dict, term):
 
 def ts_to_dt(timestamp):
     if isinstance(timestamp, datetime.datetime):
-        logging.warning('Expected str timestamp, got datetime')
         return timestamp
     dt = dateutil.parser.parse(timestamp)
     # Implicitly convert local timestamps to UTC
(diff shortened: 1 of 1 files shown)


Qmando:
Those were originally in there to help bug fix and find locations where the conversion code was called unnecessarily. These warnings *shouldn't* ever be happening, but yeah, I'm happy to remove, as there's nothing wrong with using them in this way.

PR #1633, opened 2018-04-05
Added support for alerts for Alerta. Taken from Pull Request #848

JoseIgnacioTamayo:
Added support for alerts for Alerta alerts. 

Taken from Pull Request #848. Original code by mircopolo

Added documentation, AlertaAlerter and corresponding tests

This Alerter allows to directly build and send the JSON payload for the Alerta POST method. The main feature is that any field in the Alerta JSON can be completed by values from the match, in a similar way to CommandAlerter. If the field is missing, it is replaced by a default string.


JoseIgnacioTamayo:
Sorry... my git client is setting local merging messages.

PR #1637, opened 2018-04-06
Update ruletypes.rst - is_enabled config option

kylestlb:
`is_enabled` isn't in the docs, almost built my own until I looked at the code.


kylestlb:
@Qmando could we get this merged?

PR #1647, opened 2018-04-17
Factor out simplejson and just use json

asottile:
json is stdlib and since elastalert doesn't support python2.6 it doesn't
provide any speed benefits.  It also looks like (?) it was only being used in
a test?


asottile:
CC @Qmando

PR #1666, opened 2018-04-28
Added the possibility to choose a document field in a spike rule

snuids:
@Qmando 

Hi,

I required the possibility to trigger spike rules on a particular field value of the document and not on the count of the documents. As I am lazy, and as I don't want to write a new rule using aggregations, I simply added a few tests in your spike rule code. It is not optimal but it works for me and did not create any regressions (At least for me). If you are interested in this pull request, I can fill the documentation accordingly. 

Cheers,

Arnaud


elastalert/example_rules/spiketest.yaml
```diff
@@ -0,0 +1,79 @@
...
+    query_string:
+#      query: "messages:* AND module: Skeleton"
+      query: "messages:*"
+
+# (Required)
+# The alert is use when a match is found
+alert:
+  - "stomp"
+
+stomp_hostname: app.rpinum.be
```
Qmando: @snuids  Did you mean to include this??

snuids: Ooops will change some parameters :-)

snuids: @Qmando thanks for the heads up. Server fixed. This kind of thing happens when you are doing too many things at the same time.

PR #1668, opened 2018-05-01
Add JIRA Transitions

pblasquez:
Transition a JIRA ticket from one Status to another.

### DIFF
--- a/elastalert/alerts.py
+++ b/elastalert/alerts.py
@@ -783,8 +783,7 @@ class JiraAlerter(Alerter):
                     if self.labels:
                         for l in self.labels:
                             try:
-                                self.issue = self.client.issue(ticket)
-                                self.issue.fields.labels.append(l)
+                                ticket.fields.labels.append(l)
                             except JIRAError as e:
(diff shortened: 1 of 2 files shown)


elastalert/alerts.py
```diff
@@ -769,10 +780,26 @@ def alert(self, matches):
                         self.comment_on_ticket(ticket, match)
                     except JIRAError as e:
                         logging.exception("Error while commenting on ticket %s: %s" % (ticket, e))
+                    if self.labels:
+                        for l in self.labels:
+                            try:
+                                self.issue = self.client.issue(ticket)
```
Qmando: Doesn't this return the same object as ticket? IE, could just do ``ticket.fields.labels.append(l)``?

pblasquez:
Thanks!

Re: ticket.fields.labels.append(1) - this is the better way, thank you for pointing it out. I couldn't find a good example in this particular function and I'm still getting used to this codebase.  Currently testing it in my environment to confirm, will submit soon.

I'll look into why the tests are failing and will submit a fix for that too.

Issue #1677, opened 2018-05-05
is_enabled not work with reload.


xubo-fei:
We tried set is_enabled to false for rule without restart elastalert. however, the rule sill take effect.

https://github.com/Yelp/elastalert/blob/master/elastalert/elastalert.py#L982

It should remove the disabled rule from the active rules, not just continue.

kylestlb:
@Qmando is this expected? Does is_enabled require a restart of elastalert process?

Qmando:
@xubo-fei 
Thanks for reporting this bug.

@kylestlb 
Not expected. I'll fix this ASAP. One workaround right now is instead of adding ``is_enabled: true``, rename the file whatever.yaml.disabled or similar.

hottwister:
Same issue. Still not patched?

rafaelvbsalgado:
Still not fixed...

daichi703n:
Hi, @hottwister , @rafaelvbsalgado 
I have same issue and fixed it my own. It works for me as expected. Could you try this changes on your environment?
Changes are: https://github.com/daichi703n/praeco/compare/1481254496fc7f9b8e10433c2c943218b827c97c...b85e1322c1e4866f76c1bdb78f657ec0034df939

Hi, @Qmando 
I don't create PR yet, because my fixes are poor and redundant... Would you please re-consider about handling of disabled_rule?
I'm using [praeco](https://github.com/ServerCentral/praeco) produced by ServerCentral. In it, disabling rule is so easy and useful, therefore I'm expecting proper handling.

daichi703n:
@hottwister , @rafaelvbsalgado Ping

hottwister:
Sorry, too scary testing it in prod env :)

daichi703n:
@hottwister Don't you have testing/development env?
I agree with your scaring. On my env, only not mission-critical alerting is working with my code.

I also wait for a reply from the developer.

nsano-rururu:
@daichi703n 

Is there a problem with this issue when applied in a real environment? If there is no problem,
 I think you should issue a PR.
 I don't think it's a problem that the fix is redundant, 
but it's important to fix the problem. 
Any problems with the code should be pointed out in the review, 
and if you need to add test code, Travis CI's automated testing will fail.
I don't think you need to worry too much.

nsano-rururu:
I forgot to ask if this issue has been resolved or unsolved

PR #1683, opened 2018-05-07
Fixed is_enabled on rule change

Qmando:
Fixes ``is_enabled: false`` not disabling a running rule. #1677 

Tested this manually.


kylestlb:
@Qmando any chance we can get this tagged as a patch version?

kylestlb:
@Qmando ^ Checking in to see if we can get this fix tagged as a release :)

PR #1704, opened 2018-05-17
Allows individual rule to specify a filter

stevegszabo:


Qmando:
Thanks again! Seems obvious that it should have been there, but I guess we never actually used it like this.

PR #1715, opened 2018-05-25
Limit downloaded documents when testing and use them in rules

pdscopes:
Currently the number of documents downloaded when saving documents in a test is hard coded to `10000`. This pull request changes that to use a new `args.max_query_size` (this also overrides the `max_query_size` in `conf`).

Then, if the `--use-downloaded` flag is provided the downloaded documents are used when checking for matches of the rule.

### DIFF
--- a/elastalert/test_rule.py
+++ b/elastalert/test_rule.py
@@ -133,10 +133,10 @@ class MockElastAlerter(object):
                 print("top_count_key %s may be missing" % (term), file=sys.stderr)
         print('')  # Newline
 
-        # Download up to 10,000 documents to save
+        # Download up to max_query_size (defaults to 10,000) documents to save
         if args.save and not args.count:
             try:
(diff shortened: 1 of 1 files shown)


elastalert/test_rule.py
```diff
@@ -349,6 +349,17 @@ def run_rule_test(self):
             action='store',
             dest='save',
             help='A file to which documents from the last day or --days will be saved')
+        parser.add_argument(
+            '--use-downloaded',
+            action='store_true',
+            dest='use_downloaded',
+            help='Use the downloaded '
+        )
+        parser.add_argument(
```
Qmando: Can you add a default to this? I think Elasticsearch defaults to 10 , which I think is much too small for this use case

pdscopes: The default was 10000, so I'll make that the default.

PR #1716, opened 2018-05-26
Sentry integration

deric:
Added error reporting to [Sentry](https://sentry.io), when ENV variable `SENTRY_DSN` is defined.

Any Python exception or logger message with severity at least `WARNING` will be reported to Sentry (if configured).


nsano-rururu:

PR #1772, opened 2018-06-28
added search_extra_index

Qmando:
Adds a new feature to add a extra index to the previous end of each query. This is useful in the case where timestamp_field more recent than the timestamp used to generate index names.

This is a bit limited right now since it's only 1 and only on the previous end, but it should cover the most common use case.

### DIFF
--- a/elastalert/util.py
+++ b/elastalert/util.py
@@ -192,19 +192,19 @@ def format_index(index, start, end, add_extra=False):
     start -= start.utcoffset()
     end -= end.utcoffset()
     original_start = start
-    indexes = set()
+    indices = set()
     while start.date() <= end.date():
-        indexes.add(start.strftime(index))
+        indices.add(start.strftime(index))
         start += datetime.timedelta(days=1)
-    num = len(indexes)
+    num = len(indices)
     if add_extra:
-        while len(indexes) == num:
+        while len(indices) == num:
             original_start -= datetime.timedelta(days=1)
             new_index = original_start.strftime(index)
             assert new_index != index, "You cannot use a static index with search_extra_index"
-            indexes.add(new_index)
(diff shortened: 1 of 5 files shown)


elastalert/util.py
```diff
@@ -185,17 +185,22 @@ def hashable(obj):
...
+def format_index(index, start, end, add_extra=False):
     """ Takes an index, specified using strftime format, start and end time timestamps,
     and outputs a wildcard based index string to match all possible timestamps. """
     # Convert to UTC
     start -= start.utcoffset()
     end -= end.utcoffset()
-
-    indexes = []
+    original_start = start
+    indexes = set()
```
danielpops: nit indices

tests/util_test.py
```diff
@@ -175,3 +176,14 @@ def test_resolve_string(ea):
     assert resolve_string(new_style_strings[0], match) == expected_outputs[0]
     assert resolve_string(new_style_strings[1], match) == expected_outputs[1]
     assert resolve_string(new_style_strings[2], match) == expected_outputs[2]
+
+
+def test_format_index():
```
danielpops: Can you add a test for weekly indices too?

elastalert/util.py
```diff
@@ -185,17 +185,24 @@ def hashable(obj):
...
     while start.date() <= end.date():
-        indexes.append(start.strftime(index))
+        indexes.add(start.strftime(index))
         start += datetime.timedelta(days=1)
+    num = len(indexes)
+    if add_extra:
+        while len(indexes) == num:
+            original_start -= datetime.timedelta(days=1)
+            new_index = original_start.strftime(index)
```
danielpops: won't new_index == old_index if it's a weekly or monthly index?

PR #1778, opened 2018-07-03
Added option to turn off new_term raw postfix

Qmando:
- Add an option to turn off adding the .keyword/.raw postfix
- Fixed the exception for if you use terms query and have multiple fields
- Added a warning for potential mistmatch between initial query and terms query
- Updated documentation with more information

I also considered making the query_key also have an automatic .keyword/.raw postfix in general, but this changes behavior pretty drastically, and I don't want this to be a big breaking change. This is also why I made the option default to being on. I believe .keyword is added IF you are using Logstash, which I really have no idea if it's the majority or not.

### DIFF
--- a/elastalert/ruletypes.py
+++ b/elastalert/ruletypes.py
@@ -635,10 +635,16 @@ class NewTermsRule(RuleType):
             raise EAException("fields must not be an empty list")
         if type(self.fields) != list:
             self.fields = [self.fields]
-        if self.rules.get('use_terms_query') and (
-            len(self.fields) != 1 or len(self.fields) == 1 and type(self.fields[0]) == list
-        ):
+        if self.rules.get('use_terms_query') and \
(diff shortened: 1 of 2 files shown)


elastalert/ruletypes.py
```diff
@@ -635,10 +635,16 @@ def __init__(self, rule, args=None):
             raise EAException("fields must not be an empty list")
         if type(self.fields) != list:
             self.fields = [self.fields]
-        if self.rules.get('use_terms_query') and (
```
KevinHock: Super Nit: I think our preferences differ, but I found the former more slightly more readable, 
same w/r/t doing

```python
if (
    not self.rules.get('query_key').endswith('.keyword') and
    not self.rules.get('query_key').endswith('.raw') and
    self.rules.get('use_keyword_postfix', True)
):
```
below.

PR #1787, opened 2018-07-06
Add detect-secrets hook & baseline

KevinHock:


.pre-commit-config.yaml
```diff
@@ -17,3 +17,9 @@ repos:
     sha: v0.3.5
     hooks:
     -   id: reorder-python-imports
+-   repo: git://github.com/Yelp/detect-secrets
+    sha: 0.9.1
```
KevinHock: Looks wrong, but it is right

PR #1793, opened 2018-07-10
add support of compound query keies for function get_hits_terms

fang19911030:
Tested this with the rule：
```
name: xxxx-test-rule
type: frequency
num_events: 5
timeframe:
  hours: 1
alert:
  - debug
top_count_keys: ['apple']
index: pfang-test
query_key:
  - host
  - port
raw_count_keys: false
```

### DIFF
--- a/elastalert/elastalert.py
+++ b/elastalert/elastalert.py
@@ -460,26 +460,27 @@ class ElastAlerter():
 
     def get_hits_terms(self, rule, starttime, endtime, index, key, qk=None, size=None):
         rule_filter = copy.copy(rule['filter'])
-        qk_list = qk.split(", ")
-        end = None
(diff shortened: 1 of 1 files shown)


elastalert/elastalert.py
```diff
@@ -460,15 +460,27 @@ def get_hits_count(self, rule, starttime, endtime, index):
 
     def get_hits_terms(self, rule, starttime, endtime, index, key, qk=None, size=None):
         rule_filter = copy.copy(rule['filter'])
-        if qk:
+        qk_list = qk.split(", ")
```
Qmando: I think this will fail if qk is not provided, since it will be None object. We probably need to keep the ``if qk:`` if statement.

PR #1799, opened 2018-07-12
add tag support for aggregation summry text

fang19911030:
Passed all the tests

### DIFF
--- a/elastalert/alerts.py
+++ b/elastalert/alerts.py
@@ -256,9 +256,9 @@ class Alerter(object):
         return 80
 
     def get_aggregation_summary_text(self, matches):
-        text = self.rule.get('summary_prefix', '')
+        text = ''
         if 'aggregation' in self.rule and 'summary_table_fields' in self.rule:
-            summary_table_fields = self.rule['summary_table_fields']
+            text = self.rule.get('summary_prefix', '')
             if not isinstance(summary_table_fields, list):
                 summary_table_fields = [summary_table_fields]
             # Include a count aggregation so that we can see at a glance how many of each aggregation_key were encountered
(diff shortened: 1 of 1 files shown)


elastalert/alerts.py
```diff
@@ -257,6 +257,10 @@ def get_aggregation_summary_text__maximum_width(self):
 
     def get_aggregation_summary_text(self, matches):
         text = ''
+        summary_prefix = None
+        if 'summary_prefix' in self.rule and self.rule['summary_prefix']:
+            summary_prefix = self.rule['summary_prefix']
+            text = '{{{0}}}'.format(summary_prefix)
```
Qmando: I don't think we should include `{}` automatically. If brackets are needed, the user would add them into the summary_prefix option.

elastalert/alerts.py
```diff
@@ -257,6 +257,10 @@ def get_aggregation_summary_text__maximum_width(self):
 
     def get_aggregation_summary_text(self, matches):
         text = ''
+        summary_prefix = None
+        if 'summary_prefix' in self.rule and self.rule['summary_prefix']:
```
Qmando: This works fine, but you could shorten it slightly using .get with a default. If we don't add the `{}` part, it could be:
```
text = self.rule.get('summary_prefix', '')
```

elastalert/alerts.py
fang19911030: Sounds reasonable, I will change it.

elastalert/alerts.py
```diff
@@ -256,7 +256,7 @@ def get_aggregation_summary_text__maximum_width(self):
         return 80
 
     def get_aggregation_summary_text(self, matches):
-        text = ''
+        text = self.rule.get('summary_prefix', '')
```
Qmando: Just realized that this needs to go into the if statement. Otherwise we might get a lone summary_prefix.
```
text = ''
if 'aggregation' in self.rule and .....
    text = self.rule.get('summary_prefix', '')
```

fang19911030: Good observation！

PR #1834, opened 2018-07-27
add auth support for notification email

fang19911030:
Passed all tests.

### DIFF
--- a/elastalert/alerts.py
+++ b/elastalert/alerts.py
@@ -11,8 +11,6 @@ import warnings
 from email.mime.text import MIMEText
 from email.utils import formatdate
 from HTMLParser import HTMLParser
-from smtplib import SMTP
-from smtplib import SMTP_SSL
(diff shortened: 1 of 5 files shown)


fang19911030:
smtp_auth_file for non－alerter email（notification email） should be put in the config file.

Qmando:
``smtp_auth_file`` appears to only be referenced in Alerter class, __init__ and collect_config:
```
if self.rule.get('smtp_auth_file'):
            smtp_config['user'] = self.user
            smtp_config['password'] = self.password
```

So the user/password would technically work if entered directly into the rule, but not if smtp_auth_file is used.

fang19911030:
Sorry.  I did't say it very clearly. For notification_email, I think the user and password of this email account could be put in the config file, if this account need authentication. If we wanna move the get_account to util, it may need more complex changes. Now, it seems only work to read email_config from rule folder.

PR #1869, opened 2018-08-28
Pull down all fields for frequency use_terms_query action.

chngtrn:
Frequency rule use_terms_query doesn't pull down all the fields even though "include" was used. I've modified the code so that it'll pull down all the fields. Please review and see if this can be merged and put into v0.1.35. Thanks.

### DIFF
--- a/elastalert/elastalert.py
+++ b/elastalert/elastalert.py
@@ -49,7 +49,6 @@ from util import ts_now
 from util import ts_to_dt
 from util import unix_to_dt
 
-
 class ElastAlerter():
     """ The main ElastAlert runner. This class holds all state about active rules,
     controls when queries are run, and passes information between rules and alerts.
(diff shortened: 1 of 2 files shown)


Qmando:
I don't want to accept this as it is, because it kind of beats the point of `use_terms_query`, which is to *not* query for every single document. This is effectively turning it back into a normal query, except with the bucket counts already available.

One area where we could improve on, or model this around, is how `top_count_keys` works. In that case, additional queries are made only after a match is generated. This way, we avoid unnecessarily making any queries. 

The current problem with top_count_keys is that it's not very flexible. You can only get the terms aggregation numbers for other fields, not full documents. People might want to be able to attach full documents after a match, similar to what you are doing.

chngtrn:
I did his because the result of my query returns multiple records but I want to alert for every record and include all the fields in each record in the alert. Also, I only want to alert if the unique query_key happened at least N times (num_events).

How would you suggest me accomplish that without using terms query?

Qmando:
Why do you need to use `use_terms_query` at all? Just remove it and it will do what you want.

chngtrn:
I tried that but it didn’t seem to work for my use case. Let me try again. Thanks.

Qmando:
`it’ll only send out one alert for every run`
Try setting
```
realert:
  minutes: 0
```

PR #1928, opened 2018-09-26
feat: enhance PagerDuty alert support for dynamic field values

aidan-:
The recent changes to the PagerDuty alert added support for their v2 API and with it a number of additional fields.  The fields that were added are a lot more useful if you can use field values from the rule or the actual match payload.  This is hinted at in the existing documentation:
>``pagerduty_v2_payload_source``: Sets the source of the event, preferably the hostname or fqdn.

but there is currently no mechanism to do so.

This PR adds that support while maintaining backwards compatibility for existing configurations and updates the documentation to match.


kskewes:
Nice!

aidan-:
@Qmando Any change of this getting merged?

PR #1950, opened 2018-10-09
Fix starttime in subsequent runs when allow_buffer_time_overlap is true

sureshkumarpanneerselvam:


sureshkumarpanneerselvam:
This change is to fix the issue - starttime looks incorrect from the second run onwards, if allow_buffer_time_overlap is set to true
https://github.com/Yelp/elastalert/issues/1951

sureshkumarpanneerselvam:
@Qmando Can you please review the changes and provide your comments?

sureshkumarpanneerselvam:
Thanks @Qmando for accepting the changes. Can you please let me know when will be the next release of Elastalert?

PR #1955, opened 2018-10-10
Timeout support for Slack

### LINKED ISSUE #1954: Elastalert frozen on sending Slack alert
We have seen a few times that Elastalert "0.1.18" completely stops working after a loss of network connectivity.

With some investigation we traced this back to the alerting to Slack, which seems to get stuck waiting for a response from the unreachable slack servers, and never times out. After that it seems that Elastalert does not process any other rule.

I've looked for configuration options to set a timeout, and at  the code but as far as I can see the `SlackAlerter:alert` method does not set `timeout` in the call to `requests.post(...)`. This seems unchanged in current master branch.

Is there any known workaround/fix ?

( Python 2.7.5 on Centos7 in a virtenv, elastalert-0.1.18, requests-2.19.1 , packaged with https://github.com/amine7536/elastalert-rpm )

kaplun:
This closes #1954

### DIFF
--- a/docs/source/ruletypes.rst
+++ b/docs/source/ruletypes.rst
@@ -1479,10 +1479,10 @@ Optional:
 
 ``opsgenie_recipients``: A list OpsGenie recipients who will be notified by the alert.
 ``opsgenie_recipients_args``: Map of arguments used to format opsgenie_recipients.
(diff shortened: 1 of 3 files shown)


kaplun:
@ballestr this fixes your issue by introducing a configurable default timeout of 10 seconds (similar to the generic HTTP POST request alerter.

PR #1985, opened 2018-11-01
add: custom timeframe testing support

### LINKED ISSUE #1978: Add support to test rules with custom timeframe
Currently, for testing any rule the _elastalert-test-rule_ script runs ElastAlert over the last 24 hours and prints out any alerts that would have occurred. It would be better if we don't ignore the _timeframe_ given in the rule's yml and run the test for only that period. If no _timeframe_ is given in the rule, then it may default to 24 hours.
see [Testing Your Rule](https://elastalert.readthedocs.io/en/latest/ruletypes.html#testing-your-rule)

abhishekjiitr:
Fixes #1978 
Currently, for testing any rule the elastalert-test-rule script runs ElastAlert over the last 24 hours and prints out any alerts that would have occurred.
see [Testing Your Rule](https://elastalert.readthedocs.io/en/latest/ruletypes.html#testing-your-rule) 
This PR fixes this. The timeframe given in the rule is read and test is run for only that period. If days is given as command line argument, it overrides the timreframe given in rule. If no timeframe is given in the rule and no command line argument of days is given, it defaults to 24 hours.
Also, default value of days(command line arg) is set to 0 instead of 1

### DIFF
--- a/elastalert/test_rule.py
+++ b/elastalert/test_rule.py
@@ -274,7 +274,16 @@ class MockElastAlerter(object):
                     self.handle_error("%s is not a valid ISO8601 timestamp (YYYY-MM-DDTHH:MM:SS+XX:00)" % (args.start))
                     exit(1)
             else:
-                starttime = endtime - datetime.timedelta(days=args.days)
+                # if days given as command line argument
+                if args.days > 0:
+                    starttime = endtime - datetime.timedelta(days=args.days)
+                else:
+                    # if timeframe is given in rule
+                    if 'timeframe' in rule:
+                        starttime = endtime - datetime.timedelta(seconds=rule['timeframe'].total_seconds() * 1.01)
+                    # default is 1 days / 24 hours
+                    else:
+                        starttime = endtime - datetime.timedelta(days=1)
 
         # Set run_every to cover the entire time range unless count query, terms query or agg query used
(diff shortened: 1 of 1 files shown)


Qmando:
I like this idea, just have one comment. There's a couple places where the rules are set up with conditions on "time elapsed > timeframe" instead of "time elapsed >= timeframe", so running *exactly* the timeframe might actually cause some unexpected behavior.

Maybe we can do something like timeframe * 1.1?

abhishekjiitr:
@Qmando I agree, maybe you are pointing to the code [here](https://github.com/Yelp/elastalert/blob/6517af52abcda2f96b6ea1aef4c039518bdf0c10/elastalert/ruletypes.py#L931). I am thinking of a smaller factor like **1.01** / **1.001**, because if someone wants to run a rule for a 24 hours timeframe. His rule may actually run for 2.4 hours more if we choose the factor to be **1.1** rather than 14 minutes more for **1.01**.

Qmando:
1.01 works too!

abhishekjiitr:
@Qmando , I made the changes in the PR (running rule for timeframe * 1.01) in the the case where _timeframe_ is given in rule. But, i am not sure if I should do the same for other cases, when _timeframe_ is not given in rule, and given as command-line argument or default(when even cmdline arg is not there)

Issue #1992, opened 2018-11-08
Rename prompts for creating index appropriately


davyngugi:
For a basic-auth secured instance, `elastalert-create-index` expects to send the request to a Kibana instance.

The prompts displays "**Elasticsearch**" which is deceptive 
```
Enter Elasticsearch host: 
Enter Elasticsearch port: 
```

The error trace expected when Elasticsearch host:port is used instead of Kibana is..

```
Traceback (most recent call last):
  File "/usr/local/bin/elastalert-create-index", line 11, in <module>
    load_entry_point('elastalert==0.1.36', 'console_scripts', 'elastalert-create-index')()
  File "/usr/local/lib/python2.7/dist-packages/elastalert/create_index.py", line 129, in main
    esversion = es.info()["version"]["number"]
  File "/usr/local/lib/python2.7/dist-packages/elasticsearch/client/utils.py", line 76, in _wrapped
    return func(*args, params=params, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/elasticsearch/client/__init__.py", line 241, in info
    return self.transport.perform_request('GET', '/', params=params)
  File "/usr/local/lib/python2.7/dist-packages/elasticsearch/transport.py", line 349, in perform_request
    data = self.deserializer.loads(data, headers_response.get('content-type'))
  File "/usr/local/lib/python2.7/dist-packages/elasticsearch/serializer.py", line 79, in loads
    raise SerializationError('Unknown mimetype, unable to deserialize: %s' % mimetype)
elasticsearch.exceptions.SerializationError: Unknown mimetype, unable to deserialize: text/html
```

### Suggested Fix
Rename fields to appropriately read
```
Enter Kibana host: 
Enter Kibana port: 
```
Ref:
https://github.com/Yelp/elastalert/blob/faf372f452c04c0fe1d336b7691f7187eda2b41e/elastalert/create_index.py#L84

PR #1997, opened 2018-11-12
 FIx: (issue #1992) Rename prompts for creating index 

letmerecall:


Qmando:
This is not correct. Kibana is a separate front end for Elasticsearch.

PR #1998, opened 2018-11-13
Add slack_title option

watsonlu:
As far as I can tell, slack_title_link doesn't actually work without a slack_title set.

This PR adds the ability to pass in a slack_title so that feature works.


watsonlu:
just realized that slack_alert_fields can do this instead

watsonlu:
nevermind, I read that wrong

PR #2007, opened 2018-11-23
change silence to elastalert_silence

abhishekjiitr:
The index in elastalert is stored as `elastalert_silence` instead of `elastalert_silence`. This is confusing if the user is looking for the exact index names which elastalert uses


abhishekjiitr:
This is applicable to new ES versions i.e. ES 6+

abhishekjiitr:
@Qmando, can you review?

Issue #1363, opened 2017-09-28
constant query window?


sirudog:
my config.yaml:
  
```rules_folder: /opt/rules
  run_every:
    seconds: 10
  buffer_time:
    minutes: 2
  es_host: xyz.org
  es_port: 9200
  writeback_index: elastalert_status
  alert_time_limit:
    days: 1
```  
my rule.yam:

  ```es_host: xyz.org
  es_port: 9200
  name: My errors
  type: any
  index: errors-%Y.%m.%d
  use_strftime_index: true
  timestamp_field: timestamp
  include: ["timestamp", "logLevel", "serviceName", "logContext", "logMessage", "source"]
  filter: []
  timeframe:
    minutes: 2
  realert:
    minutes: 0
  alert:
  - "hipchat"
  ```
My indices named ```errors-%Y.%m.%d``` contain only errors. I want to be notified of every single error in hipchat, hence the any type, empty filter and realert set to 0.  
  
I see this behaviour after starting elastalert (taken from the elastalert_status index):
  
Time | hits | matches | starttime | endtime | queried window (based on buffer_time ???)
---- | ---- | ------- | --------- | ------- | -----------------------------------------
2017-09-25 06:15:34.136 |	0	| 0	| 2017-09-25 06:13:34.129	| 2017-09-25 06:15:34.129 | 2m
2017-09-25 06:15:44.148	| 0	| 0	| 2017-09-25 06:13:34.129	| 2017-09-25 06:15:44.139 | 2m 10s 
2017-09-25 06:15:54.165	| 0	| 0	| 2017-09-25 06:15:34.129	| 2017-09-25 06:15:54.157 | 20s
2017-09-25 06:16:04.178	| 0	| 0	| 2017-09-25 06:15:34.129	| 2017-09-25 06:16:04.168 | 30s
2017-09-25 06:16:14.172	| 0	| 0	| 2017-09-25 06:15:34.129	| 2017-09-25 06:16:14.165 | 40s
2017-09-25 06:16:24.182	| 0	| 0	| 2017-09-25 06:15:34.129	| 2017-09-25 06:16:24.174 | 50s
2017-09-25 06:16:34.194	| 0	| 0	| 2017-09-25 06:15:34.129	| 2017-09-25 06:16:34.186 | 1m
2017-09-25 06:16:44.204	| 0	| 0	| 2017-09-25 06:15:34.129	| 2017-09-25 06:16:44.195 | 1m 10s
[...]

PR #2031, opened 2018-12-14
Fixed a bug causing buffer_time to sometimes be ignored

Qmando:
Fixes the issue reported in #1363

While running in normal (not use_count/terms_query or metric agg) mode, will sometimes start ignoring the buffer_time when making queries and prevent any overlap. This occurs when there is a slight delay and elastalert splits a segment into two pieces, on the subsequent run, it will use the starttime of the 2nd query.

For example, if the buffer_time is set to 45 minutes, and run 0 took an extra 1 minute to run
```
Run 1 - at 12:46  
 - Query 1a 12:00 to 12:45
 - Query 1b 12:45 to 12:46
Run 2 - at 12:47
 - Query 1a 12:45 - 12:47
```

This is because there's a code path in `set_starttime` where it's never set at all, and the previous run is used.

This was introduced all the way back in 2017 by in #846

```
-            elif 'previous_endtime' in rule and rule['previous_endtime'] < buffer_delta:
-                rule['starttime'] = rule['previous_endtime']
+            elif 'previous_endtime' in rule:
+                if rule['previous_endtime'] < buffer_delta:
+                    rule['starttime'] = rule['previous_endtime']
+                self.adjust_start_time_for_overlapping_agg_query(rule)
            else:
                rule['starttime'] = buffer_delta
```

This also removes the `allow_buffer_time_overlap` "feature" which was added when someone noticed this bug but probably thought it was intentional and I didn't realize it was actually a bug fix either.

### DIFF
--- a/elastalert/elastalert.py
+++ b/elastalert/elastalert.py
@@ -689,7 +689,6 @@ class ElastAlerter():
 
     def set_starttime(self, rule, endtime):
         """ Given a rule and an endtime, sets the appropriate starttime for it. """
-
         # This means we are starting fresh
         if 'starttime' not in rule:
             if not rule.get('scan_entire_timeframe'):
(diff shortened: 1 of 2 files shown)


elastalert/elastalert.py
```diff
@@ -715,12 +714,9 @@ def set_starttime(self, rule, endtime):
             if 'minimum_starttime' in rule and rule['minimum_starttime'] > buffer_delta:
                 rule['starttime'] = rule['minimum_starttime']
             # If buffer_time doesn't bring us past the previous endtime, use that instead
-            elif 'previous_endtime' in rule:
-                if rule['previous_endtime'] < buffer_delta:
-                    rule['starttime'] = rule['previous_endtime']
-                    self.adjust_start_time_for_overlapping_agg_query(rule)
-                elif rule.get('allow_buffer_time_overlap'):
```
danielpops: is it safe to drop this functionality?

Qmando: I'm not sure I would call this functionality. `allow_buffer_time_overlap` is just a toggle to fix this bug. It's possible someone has started relying on this bug's behavior, but that seems unlikely.

PR #2038, opened 2018-12-18
Put quotes around aggregate_id search

Qmando:
Add quotes to the aggregate_id search. This prevents it from accidentally picking up incorrect alerts if the _ids share common patterns. Note that this *only* occurs if the mapping is applied incorrectly, as aggregate_id is supposed to be non-analyzed. However, we may as well preemptively prevent these errors because it's as simple as adding quotes.


nsano-rururu:

PR #2052, opened 2018-12-28
Add new Alert type for Spikes in Single Metric Aggregations

dylanjf:
An alert which fires off of rises/drops in the average value of a metric (as well as any other aggregation supported by the Metric Aggregation rule)

This piggybacks off of all of the work done in Spike alerts (which let us consider relative changes but only of counts) and Metric Aggregation alerts (which let us consider aggregated values but only against predefined thresholds)

In addition, I added support for aggregation of scripted metrics.  This is currently only supported in the new Spike Aggregation alert but could be easily extended to Metric Aggregations as well.


Qmando:
Awesome work!!

PR #2080, opened 2019-01-17
Resend update config option

thommor:
Added a config option "resend_update"

When True this config option will make ElastAlert update previous ES entries when retrying an alert rather than deleting the old alert and creating a new one.

Will perform update based on index and id.


nsano-rururu:
Test are failing

thommor:
The tests failed due to an error with the slack tests which from which I remember had nothing to do with this PR.

I gave up on this PR being merged a long time ago and have also moved away from ElastAlert due to the lack of upkeep on this repo

PR #2094, opened 2019-01-25
Fixes FrequencyRule.add_data to run check_for_match on all keys seen

pdscopes:
`FrequencyRule.add_data` runs a final `self.check_for_match(key, end=True)` after it has added all the data provided.

There are subclasses such as `FlatlineRule` that only checks for matches when `end=True`. Currently this means that if more than one key was found in the given data that only the last key found would actually be checked.

This fix stores a set of keys found in `data` and runs `self.check_for_match(key, end=True)` on all found keys not just the last key found.

### DIFF
--- a/elastalert/ruletypes.py
+++ b/elastalert/ruletypes.py
@@ -244,6 +244,7 @@ class FrequencyRule(RuleType):
         else:
             qk = None
 
+        keys = set()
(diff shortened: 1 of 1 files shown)


nsano-rururu:
Test are failing

PR #2096, opened 2019-01-27
Add alert_text_header and alert_text_footer options

johnsusek:
When using aggregations, any alert_text gets repeated for every match. So, if you wanted to add some text just at the beginning or end of the aggregation alert, you cannot do this. I added the options alert_text_header and alert_text_footer to allow for this. Now you can add some text to the beginning or end of aggregation alerts.


Salaander:
I actually just preparing a PR that makes it possible to use jinja templating to render the alert messages and I created a new match string formatter that can accept aggregated matches and pass it through jinja so you can have all the display logic abstracted to a template.
The reason I'm writing it here is that when I push it I would like to get some suggestion on how to make it better and conform to be able to accept it. We are now running this version locally and it makes formatting alerts much easier in my opinion. What do you think?

johnsusek:
https://github.com/ServerCentral/praeco/issues/67

nsano-rururu:

PR #2112, opened 2019-02-02
Fixes #2110 Logging inconsistencies fixed in alerts.py

### LINKED ISSUE #2110: Logging issues
I also noticed that in alerts.py there is an `import logging` that was in the _Initial commit_ yet it is not set up and only used 7 times: 6 times in *JiraAlerter* and once in *CommandAlerter*.
In `JiraAlerter.alert()` it is even used in the same function where `elastalert_logger` is also used.

I just wanted to ask if anybody else noticed these issues and to ask if you would be cool to remove the `import logging` and the 7 occurrences should be swapped to `elastalert_logger` references.
For some reason I cannot even see the logs from `alerter.py` even is they are called on the `elastalert_logger` instance and `--verbose` is set but I am debugging it.

Salaander:
Removed the import of the logging lib and all the usages since they were not used properly and hid some errors. Changed the previous logger invocations to the proper logger `elastalert_logger`.

### DIFF
--- a/elastalert/alerts.py
+++ b/elastalert/alerts.py
@@ -2,7 +2,6 @@
 import copy
 import datetime
 import json
-import logging
 import os
 import subprocess
 import sys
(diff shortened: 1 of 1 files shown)


nsano-rururu:
@Salaander

There is a description that seems to be applicable in the following file, but does it mean that the same correction is necessary?

elastalert/elastalert.py
elastalert/kibana_discover.py
elastalert/loaders.py
elastalert/opsgenie.py
elastalert/util.py

nsano-rururu (approved):

PR #2121, opened 2019-02-06
adding env var variable resolution for elasticsearch passwords

adamenger:
Hi there,

I have a situation where I'd like to have multiple rules reach out to different ES clusters.  This works with Elastalert currently, but unfortunately that means I have to maintain a web of credentials throughout a bunch of different rule files. We want to avoid committing the credentials, so right now we're using `sed` to replace placeholder values in the files before they're deployed. 

I think an environment variable for `es_username` and `es_password` would provide us a much more elegant way to have multiple passwords loaded in a single elastalert instance. 

For example, this would allow us to have two very similar rules that use two different credentials:

*example rule 1*


nsano-rururu:

PR #2113, opened 2019-02-04
Exponential resend

thommor:
Added timedelta conversion for `exponential_realert` in global config

This PR closes #2098

Issue #2098, opened 2019-01-29
Use realert error!!!


Hello-Linux:
To Reduce the frequency of repeated alarms,I set config in config.yaml
```
query_key:
  - name

realert:
  minutes: 10

exponential_realert:
  hours: 1
```
but when I send alert message the error show:
"ERROR:root:Traceback (most recent call last):
  File "/scripts/elastalert/elastalert/elastalert.py", line 1169, in run_all_rules
    num_matches = self.run_rule(rule, endtime, self.starttime)
  File "/scripts/elastalert/elastalert/elastalert.py", line 895, in run_rule
    next_alert, exponent = self.next_alert_time(rule, silence_cache_key, ts_now())
  File "/scripts/elastalert/elastalert/elastalert.py", line 1904, in next_alert_time
    if wait >= rule['exponential_realert']:
TypeError: can't compare datetime.timedelta to dict"        why??????????????When I remove these configurations, the alarm is successful

thommor:
We tested it and it seems to work as expected for us.

Could it be that your yaml indentation is slightly off?

It should be like:
```yaml
exponential_realert:
  hours: 1
```

The same would apply for your realert as well.

Qmando:
You need to put those settings into individual rules instead of the global config.yaml. I think this qualifies as a bug since you're supposed to be able to put defaults in the global config, but in this case, the code that converts it to a timedelta object doesn't run unless it's in the rule yaml.

PR #2137, opened 2019-02-15
Correctly set `realert`, `aggregation`, `query_delay` from the config

kiwiz:
Hi! It seems that the `query_delay` field isn't correctly inherited from the config file. Trying to set `query_delay: {'minutes': 1}` in `elasticsearch.yml` results in the following exception:
```
Traceback (most recent call last):
  File "/usr/local/bin/elastalert", line 11, in <module>
    load_entry_point('elastalert==0.1.38', 'console_scripts', 'elastalert')()
  File "/usr/local/lib/python2.7/dist-packages/elastalert/elastalert.py", line 1919, in main
    client.start()
  File "/usr/local/lib/python2.7/dist-packages/elastalert/elastalert.py", line 1100, in start
    self.run_all_rules()
  File "/usr/local/lib/python2.7/dist-packages/elastalert/elastalert.py", line 1162, in run_all_rules
    endtime = ts_now() - delay
TypeError: unsupported operand type(s) for -: 'datetime.datetime' and 'dict'
```
Inspecting the code, it looks like this [line](https://github.com/Yelp/elastalert/blob/master/elastalert/config.py#L216) is setting `query_delay` to the `dict` instead of converting it to a `datetime`. This PR should fix the issue for all the relevant `timedelta` fields.

### DIFF
--- a/elastalert/config.py
+++ b/elastalert/config.py
@@ -212,6 +212,10 @@ def load_options(rule, conf, filename, args=None):
         raise EAException('Invalid time format used: %s' % (e))
 
     # Set defaults, copy defaults from config.yaml
+    td_fields = ['realert', 'aggregation', 'query_delay']
+    for td_field in td_fields:
+        if td_field in base_config:
+            rule.setdefault(td_field, datetime.timedelta(**base_config[td_field]))
     for key, val in base_config.items():
         rule.setdefault(key, val)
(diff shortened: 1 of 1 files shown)


kiwiz:
Opps, just took a look at open issues and it seems like there's already some conversation about this: #2098 #2113

PR #2151, opened 2019-02-22
Add support for `min_doc_count` to terms queries

kiwiz:
This allows `use_terms_query` to specify a threshold for individual buckets. I noticed that the new `metric_aggregation` rules also allow setting `min_doc_count`, so lemme know if this is the wrong approach.


kiwiz:
Sorry for the noise, change should be ready now.

Issue #1919, opened 2018-09-20
Using Elastalert to notify us when "Service are back online"


Purfakt:
I created a `flatline` type rule that sends an email when there is no more incoming message from a triplet that represent one of our service (this one works) : 

```
name: Service Down
type: flatline

index: 'ourindex'

timestamp_field: 'ourtimestamp'
timestamp_type: unix_ms

query_key: ["s_serviceInfo-instanceId", "s_serviceInfo-replicaId", "s_serviceInfo-serviceName"]

realert:
  hours: 1

filter:
  - query_string:
      query: "s_qs_item-name: machine.cpu"

timeframe:
  minutes: 5

threshold: 1

alert:
  - email
from_addr: "alert@domain.com"
email: "me@domain.com"

alert_subject: "Service down on {0}"
#key represent the query key
alert_subject_args: 
  - "key"

alert_text: |
    Service not answering on {0} at {1}
alert_text_args: 
  - "key"
  - "ourtimestamp"

alert_text_type: alert_text_only
```

Now I need to create an alarm to notify us when this service is back online.
I thought I'd be smart with creating a `new_term` with `term_window_size` that matches the `timeframe` of the `flatline` alarm: 

```
name: Service up

type: new_term

index: 'ourindex'

timestamp_field: 'ourtimestamp'
timestamp_type: unix_ms

fields: 
  - "s_serviceInfo-instanceId" 
  - "s_serviceInfo-replicaId" 
  - "o_serviceInfo-propertiesMessageId"

query_key: ["s_serviceInfo-instanceId", "s_serviceInfo-replicaId", "s_serviceInfo-serviceName"]

realert:
  hours: 1

filter:
  - query_string:
      query: "s_qs_item-name: machine.cpu"

terms_window_size:
  minutes: 5

window_step_size:
  minutes: 1

alert:
  - email
from_addr: "alert@domain.com"
email: "me@domain.com"

alert_subject: "Service up on {0}"
alert_subject_args: 
  - "s_serviceInfo-instanceId"

alert_text: |
    Service is now up and running on {0}, {1} at {2}
alert_text_args: 
  - "s_serviceInfo-instanceId"
  - "s_serviceInfo-replicaId"
  - "ourtimestamp"

alert_text_type: alert_text_only
```

Obviously, I'm either misunderstanding something or at least misusing it because the first alarm is working great on all services but the second one only triggers when it is first added to the rules folder. There is 0 match and the rule isn't silenced.

[...]

PR #2153, opened 2019-02-28
Es6 writebackindex fix

damioune123:
Hi I'd like to introduce some features I've added to elastalert:
1) As when using "use_query_terms" in flatline rule, it is only possible to aggregate on one field, I've added the new feature of elasticsearch v6 to use composite key ( https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-composite-aggregation.html ).
As the composite key may cause problem when using none ".keyword" fields with unindexed fields, you can configure the rule to use or not the ".keyword" field suffix with the following rule config :
" use_keyword_suffix: false/true " 

2) As related to following issue : https://github.com/Yelp/elastalert/issues/1919 , it was not possible to detect whenever a "service" is back again when using flatline rule. As the state kept by elastalert in its writeback index doesn't change whenever the service is detected again. I added the possibility to write to elastalert writeback_index whenever a service is detected again. 2 more config rules are now possible in a flatline rule : 
writeback_events_up_enabled: true => whether to write in elastalert writeback index whenever a service is detected again
event_up_rule_name_suffix: "_UP" => the suffix of the "rule_name" field that will be used when inserting in writeback_index 
For example when creating a flatline rule alert named "Service Down", elastalert will write in writeback index a doc with the field rule_name set to "Service Down", a same doc will be written with the rule_name "Service_Down_UP" whenever the service is detected again. 
So for instance you can create a frequency rule watching elastalert writeback with: 
filter:
- query_string:
      query: "rule_name: Service\\ Down_UP"
and it will notice you of services up again

3) As it is now possible to use_query_terms with composite keys, I've added the possibility to run the query at once (not sliced by buffer_time/run_every), very useful when combined with the new feature (not mine, but I've merged it here) of limit_execution where you can set cron for each rule.
rule config : 
query_all_timeframe_at_once: true
Very useful when running a rule every hour for instance but your sliced query has a size of 1 min, instead of doing 60 sequential queries of 1 min, you can query your whole time frame at once.

4) use_timeframe_as_max_limit: true/false
This parameter allows to limit the starttime limit of a query. Imagine you want to check if services are up from 7:00 to 19:00. When using limit_execution cron, the rule will wake up at 7:00 and detect that is hasn't run for 12 hours, so it will queries the whole timeframe "gap" from 19:00 to 7:00. With this parameter , the start_time of the query will be the max between now - timeframe and the last query run. 

5) Small bug fix with limit_execution, as the  "has_run_once" bool var may sometimes never be set and so the rule will never run


thommor:
You appear to be trying to push the beta branch into master.

I would also suggest splitting this into multiple PRs (one for each feature) and change them to go to the beta branch

PR #2168, opened 2019-03-13
ES6 writeback index fix + extra features

damioune123:
Hi I'd like to introduce some features I've added to elastalert:
1) As when using "use_query_terms" in flatline rule, it is only possible to aggregate on one field, I've added the new feature of elasticsearch v6 to use composite key ( https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-composite-aggregation.html ).
As the composite key may cause problem when using none ".keyword" fields with unindexed fields, you can configure the rule to use or not the ".keyword" field suffix with the following rule config :
" use_keyword_suffix: false/true " 

2) As related to following issue : https://github.com/Yelp/elastalert/issues/1919 , it was not possible to detect whenever a "service" is back again when using flatline rule. As the state kept by elastalert in its writeback index doesn't change whenever the service is detected again. I added the possibility to write to elastalert writeback_index whenever a service is detected again. 2 more config rules are now possible in a flatline rule : 
writeback_events_up_enabled: true => whether to write in elastalert writeback index whenever a service is detected again
event_up_rule_name_suffix: "_UP" => the suffix of the "rule_name" field that will be used when inserting in writeback_index 
For example when creating a flatline rule alert named "Service Down", elastalert will write in writeback index a doc with the field rule_name set to "Service Down", a same doc will be written with the rule_name "Service_Down_UP" whenever the service is detected again. 
So for instance you can create a frequency rule watching elastalert writeback with: 
filter:
- query_string:
      query: "rule_name: Service\\ Down_UP"
and it will notice you of services up again

3) As it is now possible to use_query_terms with composite keys, I've added the possibility to run the query at once (not sliced by buffer_time/run_every), very useful when combined with the new feature (not mine, but I've merged it here) of limit_execution where you can set cron for each rule.
rule config : 
query_all_timeframe_at_once: true
Very useful when running a rule every hour for instance but your sliced query has a size of 1 min, instead of doing 60 sequential queries of 1 min, you can query your whole time frame at once.

4) use_timeframe_as_max_limit: true/false
This parameter allows to limit the starttime limit of a query. Imagine you want to check if services are up from 7:00 to 19:00. When using limit_execution cron, the rule will wake up at 7:00 and detect that is hasn't run for 12 hours, so it will queries the whole timeframe "gap" from 19:00 to 7:00. With this parameter , the start_time of the query will be the max between now - timeframe and the last query run. 

5) Small bug fix with limit_execution, as the  "has_run_once" bool var may sometimes never be set and so the rule will never run


Qmando:
Thanks for doing all of this work!

elastalert/elastalert.py
```diff
@@ -146,6 +146,7 @@ def __init__(self, args):
         self.from_addr = self.conf.get('from_addr', 'ElastAlert')
         self.smtp_host = self.conf.get('smtp_host', 'localhost')
         self.max_aggregation = self.conf.get('max_aggregation', 10000)
+        self.limit_execution_margin = self.conf.get('limit_execution_margin', 10)
```
Qmando: This doesn't look like it's referenced anywhere

elastalert/elastalert.py
```diff
@@ -255,17 +256,25 @@ def get_query(filters, starttime=None, endtime=None, sort=True, timestamp_field=
...
+        fields = field.split(",")
+        for field in fields:
+            if use_keyword_postfix:
+                agg_query['counts']['composite']['sources'].append({field: {"terms": {"field": field+".keyword"}}})
+            else:
+                agg_query['counts']['composite']['sources'].append({field: {"terms": {"field": field}}})
+
         if not five:
-            query_element['filtered'].update({'aggs': {'counts': {'terms': {'field': field, 'size': size}}}})
+            query_element['filtered'].update({'aggs': agg_query})
```
Qmando: Will this work on older versions?

elastalert/elastalert.py
```diff
@@ -657,6 +669,12 @@ def run_query(self, rule, start=None, end=None, scroll=False):
                 rule_inst.add_aggregation_data(data)
             else:
                 rule_inst.add_data(data)
+        if isinstance(rule['type'], FlatlineRule):
+            if "writeback_events_up_enabled" in rule_inst.rules and rule_inst.rules["writeback_events_up_enabled"] and len(rule_inst.events_up) !=0:
```
Qmando: Can you replace `"writeback_events_up_enabled" in rule_inst.rules and rule_inst.rules["writeback_events_up_enabled"]` with `rule_inst.get("writeback_events_up_enabled")` ?

Also add a space between != and 0

nsano-rururu:
Test are failing

nsano-rururu:

PR #2176, opened 2019-03-20
Added version 7 support

stevegszabo:
I started testing elastalert with ES 7, has anyone else started working on this?

### DIFF
--- a/elastalert/config.py
+++ b/elastalert/config.py
@@ -28,7 +28,7 @@ from util import unix_to_dt
 from util import unixms_to_dt
 
 # schema for rule yaml
-rule_schema = jsonschema.Draft4Validator(yaml.load(open(os.path.join(os.path.dirname(__file__), 'schema.yaml'))))
+rule_schema = jsonschema.Draft4Validator(yaml.load(open(os.path.join(os.path.dirname(__file__), 'schema.yaml')), Loader=yaml.FullLoader))
 
 # Required global (config.yaml) and local (rule.yaml)  configuration options
 required_globals = frozenset(['run_every', 'rules_folder', 'es_host', 'es_port', 'writeback_index', 'buffer_time'])
--- a/elastalert/create_index.py
+++ b/elastalert/create_index.py
@@ -248,7 +248,15 @@ def main():
     # To avoid a race condition. TODO: replace this with a real check
     time.sleep(2)
 
-    if(elasticversion > 5):
+    if(elasticversion > 6):
+        params = {'include_type_name': 'true'}
+        es.indices.put_mapping(index=index, doc_type='elastalert', body=es_mapping, params=params)
+        es.indices.put_mapping(index=index + '_status', doc_type='elastalert_status', body=ess_mapping, params=params)
+        es.indices.put_mapping(index=index + '_silence', doc_type='silence', body=silence_mapping, params=params)
+        es.indices.put_mapping(index=index + '_error', doc_type='elastalert_error', body=error_mapping, params=params)
+        es.indices.put_mapping(index=index + '_past', doc_type='past_elastalert', body=past_mapping, params=params)
+        print('New index %s created' % index)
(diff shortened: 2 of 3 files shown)


elastalert/config.py
```diff
@@ -28,7 +28,7 @@
 from util import unixms_to_dt
 
 # schema for rule yaml
-rule_schema = jsonschema.Draft4Validator(yaml.load(open(os.path.join(os.path.dirname(__file__), 'schema.yaml'))))
+rule_schema = jsonschema.Draft4Validator(yaml.load(open(os.path.join(os.path.dirname(__file__), 'schema.yaml')), Loader=yaml.FullLoader))
```
stevegszabo: see https://github.com/yaml/pyyaml/wiki/PyYAML-yaml.load(input)-Deprecation

elastalert/create_index.py
```diff
@@ -248,7 +248,15 @@ def main():
     # To avoid a race condition. TODO: replace this with a real check
     time.sleep(2)
 
-    if(elasticversion > 5):
+    if(elasticversion > 6):
+        params = {'include_type_name': 'true'}
```
stevegszabo: see https://www.elastic.co/guide/en/elasticsearch/reference/current/removal-of-types.html

elastalert/elastalert.py
```diff
@@ -279,7 +282,10 @@ def get_index_start(self, index, timestamp_field='@timestamp'):
         """
         query = {'sort': {timestamp_field: {'order': 'asc'}}}
         try:
-            res = self.current_es.search(index=index, size=1, body=query, _source_include=[timestamp_field], ignore_unavailable=True)
```
stevegszabo: Mar 19 15:30:52 search01 elastalert[22967]: ERROR:root:Error running query: RequestError(400, u'illegal_argument_exception', u'request [/filebeat-*/_search] contains unrecognized parameter: [_source_include] -> did you mean any of [_source_includes, _source_excludes]?')

elastalert/create_index.py
stevegszabo: The include_type_name parameter in the index creation, index template, and mapping APIs will default to false. Setting the parameter at all will result in a deprecation warning.

elastalert/elastalert.py
stevegszabo: https://www.elastic.co/guide/en/elasticsearch/reference/6.0/docs-get.html

If you only need one or two fields from the complete _source, you can use the _source_include & _source_exclude parameters to include or filter out that parts you need.

vs

https://www.elastic.co/guide/en/elasticsearch/reference/7.0/docs-get.html

If you only need one or two fields from the complete _source, you can use the _source_includes and _source_excludes parameters to include or filter out the parts you need.

elastalert/create_index.py
```diff
@@ -248,7 +248,15 @@ def main():
...
+        es.indices.put_mapping(index=index, doc_type='elastalert', body=es_mapping, params=params)
+        es.indices.put_mapping(index=index + '_status', doc_type='elastalert_status', body=ess_mapping, params=params)
+        es.indices.put_mapping(index=index + '_silence', doc_type='silence', body=silence_mapping, params=params)
+        es.indices.put_mapping(index=index + '_error', doc_type='elastalert_error', body=error_mapping, params=params)
+        es.indices.put_mapping(index=index + '_past', doc_type='past_elastalert', body=past_mapping, params=params)
```
synhershko: doc_type should always be `_doc` starting 6.5 and 7

stevegszabo:
As EA depends on elasticsearch-py we should wait for [this](https://github.com/elastic/elasticsearch-py/pull/911) to merge as it will affect EA

matsgoran:
In case you would be interested, https://github.com/Yelp/elastalert/pull/2194. We are currently running these changes against our ES7.0.0-beta1 instance

PR #2181, opened 2019-03-26
Add ability to lookup into arrays

Qmando:
When using lookup_es_key, you can now access arrays. See tests for examples and usage

(It looks nicer if you remove whitespace changes from the diff view settings, I removed some blank lines)
The diff detector didn't do great here so I'll explain the the changes:
I wrapped most of the method in a while loop that tries to split the full search term by each array lookup. For each section, lookup on sub_term will proceed as before, but at the end, it tries to index into the final cursor object, and continues with the next term.

### DIFF
--- a/elastalert/util.py
+++ b/elastalert/util.py
@@ -3,6 +3,7 @@ import collections
 import datetime
 import logging
 import os
+import re
 
 import dateutil.parser
 import dateutil.tz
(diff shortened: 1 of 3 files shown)


tests/util_test.py
```diff
@@ -105,6 +105,24 @@ def test_looking_up_nested_composite_keys(ea):
...
+def test_looking_up_arrays(ea):
+    record = {
+        'flags': [1, 2, 3],
+        'objects': [
+            {'foo': 'bar'},
+            {'foo': [{'bar': 'baz'}]},
+            {'foo': {'bar': 'baz'}}
+        ]
+    }
+    assert lookup_es_key(record, 'flags[]') == 1
```
danielpops: is it just grabbing the first one? that doesn't seem super intuitive to me. I would've expected "jq-esque" behavior where this returns all the objects.

Whatever you end up choosing for the end user interface / behavior semantics, can you add documentation?

tests/util_test.py
```diff
@@ -105,6 +105,24 @@ def test_looking_up_nested_composite_keys(ea):
...
+    assert lookup_es_key(record, 'flags[1]') == 2
+    assert lookup_es_key(record, 'objects[]foo') == 'bar'
+    assert lookup_es_key(record, 'objects[1]foo[0]bar') == 'baz'
+    assert lookup_es_key(record, 'objects[2]foo.bar') == 'baz'
+    assert lookup_es_key(record, 'objects[1]foo[1]bar') is None
+    assert lookup_es_key(record, 'objects[1]foo[0]baz') is None
```
danielpops: also not sure I love `None`. What do you think about `<MISSING VALUE>` to be consistent with the behavior elsewhere?

elastalert/util.py
```diff
@@ -60,27 +61,46 @@ def _find_es_dict_by_key(lookup_dict, term):
...
 
-    while len(subkeys) > 0:
-        if not dict_cursor:
-            return {}, None
+    while term:
+        split_results = re.split(r'\[(\d?)\]', term, 1)
+        if len(split_results) == 3:
+            sub_term, index, term = split_results
+            index = index or '0'
+            index = int(index)
```
danielpops: error handling for bad values?

elastalert/util.py
```diff
@@ -60,27 +61,46 @@ def _find_es_dict_by_key(lookup_dict, term):
...
     #  {'foo.bar': {'bar': 'ray'}} to look up foo.bar will return {'bar': 'ray'}, not 'ray'
     dict_cursor = lookup_dict
-    subkeys = term.split('.')
-    subkey = ''
 
-    while len(subkeys) > 0:
-        if not dict_cursor:
-            return {}, None
+    while term:
+        split_results = re.split(r'\[(\d?)\]', term, 1)
```
KevinHock: Maybe for readability you can use the keyword for `maxsplit`, `split_results = re.split(r'\[(\d?)\]', term, maxsplit=1)`

tests/util_test.py
Qmando: This function has always returned None on failure, as it says in the description. MISSING VALUE comes from elsewhere in the code

elastalert/util.py
Qmando: Meant this to be a default, so `[]` == `[0]`. I can definitely remove this, you're right that it conflicts with `jq` behavior.

tests/util_test.py
Qmando: I've removed this example. I'd like to support jq style in the future but for now, this is sufficient. Documentation added.

PR #2211, opened 2019-04-16
Add Prometheus Metrics

abhishekjiitr:
Start exporting metrics in Prometheus format on port _9090_.

Relevant Metrics added, most of the metrics have a *rule_name*_label:
(All metrics except 6th are per rule)
1. Number of hits
2. Number of matches
3. Time taken to evaluate rule
4. Number of alerts sent
5. Number of alerts not sent
6. Total Number of elastalert_errors
7. Number of silenced alerts

### DIFF
--- a/elastalert/elastalert.py
+++ b/elastalert/elastalert.py
@@ -17,6 +17,7 @@ from socket import error
 
 import dateutil.tz
 import kibana
+import prometheus_client
 import yaml
 from alerts import DebugAlerter
 from config import get_rule_hashes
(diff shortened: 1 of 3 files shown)


abhishekjiitr:
@pdscopes @Qmando @danielpops can you take a look?

pdscopes:
Firstly, this could be converted into a script that scrapes`elastalert_alert`, `elastalert_status`, etc rather than needing to be integrated directly into ElastAlert.

Secondly, you should probably create a wrapper class and move the metrics definitions and `update_metrics` into said class.

Finally, you've not provided any way of configuring/turning off the Prometheus nor provided any documentation.

abbasalizaidi:
@abhishekjiitr I am looking for something similar to this. Would be great if it can be merged.

I will be happy to do the documentation part!

PR #2213, opened 2019-04-17
Fix an issue with nested timestamp in cardinality rules (again)

nmerdan:
There are more places in the code that use dictionary keys directly and need to be updated.


Qmando:
I dream of a day where there is no direct timestamp access. Thank you for bringing us closer to that world!

PR #2230, opened 2019-04-29
Add `max_scrolling_count` setting to limit scrolling behavior

KarolyBujtor:
Problem: when the rule query has a large result set (e.g.: more than 1 million documents) it starts to load all documents to memory which eats up all of the memory on the machine.

Solution: added the `max_scrolling_count` setting to potentially limit the number of scrolling requests.

The default `max_scrolling_count` setting is 0, which doesn't alter the current behavior.

Documentation has been updated.

### DIFF
--- a/docs/source/elastalert.rst
+++ b/docs/source/elastalert.rst
@@ -148,7 +148,11 @@ configuration.
 
 ``max_query_size``: The maximum number of documents that will be downloaded from Elasticsearch in a single query. The
 default is 10,000, and if you expect to get near this number, consider using ``use_count_query`` for the rule. If this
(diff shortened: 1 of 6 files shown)


Qmando:
Thanks for the contribution, looks good!

sherifabdlnaby:
This is very useful! Why wasn't this added to release :/

Qmando:
It was. It's in the latest release, 0.2.1.

sherifabdlnaby:
The documentation is outdated then, however, I failed to get it to work, it still scrolls the entire result :/

Qmando:
It looks like the documentation failed to update from the source. I'll double check that.

Would you mind sharing your rule configuration?

PR #2231, opened 2019-05-01
Fixed the logging property in config.yaml.example

ChoiSimba:
When uncomenting and using logging,
the level of handlers, file, logger, logger.handlers is not correct.
So, configuration error occur.

Fixed the level of problematic settings.


nsano-rururu:
Merged into elastalert2.

nsano-rururu (approved):

PR #2265, opened 2019-05-28
Allow printing and formating of related_events

agix:
When `attach_related: true` there is no way to print back the `related_events`.

By default, `_add_match_items` is now called for the current match and each of the related events.

Using `debug` alerter

```
At least 4 events occurred between 2019-05-28 19:46 CEST and 2019-05-28 19:47 CEST


----------------------------------------
@timestamp: 2019-05-28T17:47:13.134Z
_id: yjWN_2oBYegyjQkzob55
_index: test
_type: _doc
field1: 1.2.3.4
field2: fourth
num_hits: 5
num_matches: 1
type: test

----------------------------------------
@timestamp: 2019-05-28 17:47:12.496000+00:00
_id: xzWN_2oBYegyjQkznr77
_index: test
_type: _doc
field1: 1.2.3.4
field2: first
type: test

----------------------------------------
@timestamp: 2019-05-28 17:47:12.727000+00:00
_id: yDWN_2oBYegyjQkzn77i
_index: test
_type: _doc
field1: 1.2.3.4
field2: second
type: test

----------------------------------------
@timestamp: 2019-05-28 17:47:12.912000+00:00
_id: yTWN_2oBYegyjQkzoL6c
_index: test
_type: _doc
field1: 1.2.3.4
field2: third
type: test

```

It's also possible to specify a format with same syntaxe as `alert_text`

```
alert_text_related_event_text: |
  * From {0} to {1}
alert_text_related_event_text_args:
  - field1
  - field2
```

```
At least 4 events occurred between 2019-05-28 19:47 CEST and 2019-05-28 19:48 CEST

* From 1.2.3.4 to fourth
* From 1.2.3.4 to first
* From 1.2.3.4 to second
* From 1.2.3.4 to third
```

PR #2266, opened 2019-05-28
Add related_events capability to cardinality

agix:
As for `frequency` rule, it may be interesting to link `related_events` of the `cardinality` rule.

Using same name allowes common functions on `related_events` to later be used like https://github.com/Yelp/elastalert/pull/2265


agix:
Merged in https://github.com/Yelp/elastalert/pull/2585

PR #2269, opened 2019-05-30
Alerta: Add Customer in Alert Payload

abhishekjiitr:
Add customer key in Alerta alert payload derived from `alerta_customer` in rule, defaulting to `customer` if defined in rule otherwise None. The `alerta_customer` is allowed in rule syntax (`schema.yaml`) but it was not being sent in Alerta alert payload.

Given, customer is a primary attribute in Alerta, this will be useful for many people using Elastalert with Alerta.


nsano-rururu:
@abhishekjiitr

PR #2271, opened 2019-06-03
Clear scroll if not used anymore

agix:
Following this commit https://github.com/elastic/elasticsearch/commit/d27aa72b17f1ceda156f705a262ef8c5f4106c35 in elasticsearch. A scroll parallel limit of 500 is set.

This seems related to this issue : https://github.com/Yelp/elastalert/issues/2249

Elastalert create scroll for each research with default to 30s for keeping the scroll context alive. 

But it never clears the scroll if it's not used anymore

As it's really easy to open 500 scroll in 30 seconds, the scroll should be clear after usage.

### DIFF
--- a/elastalert/elastalert.py
+++ b/elastalert/elastalert.py
@@ -351,7 +351,8 @@ class ElastAlerter(object):
                     ignore_unavailable=True,
                     **extra_args
                 )
-                rule['scroll_id'] = res['_scroll_id']
+                if '_scroll_id' in res:
+                    rule['scroll_id'] = res['_scroll_id']
 
                 if self.current_es.is_atleastseven():
                     self.total_hits = int(res['hits']['total']['value'])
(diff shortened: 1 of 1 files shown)


elastalert/elastalert.py
```diff
@@ -624,7 +625,8 @@ def run_query(self, rule, start=None, end=None, scroll=False):
             pass
 
         if 'scroll_id' in rule:
-            rule.pop('scroll_id')
+            scroll_id = rule.pop('scroll_id')
+            self.current_es.clear_scroll(scroll_id=scroll_id)
```
dinesh-pratti, line 629: @Qmando For clearing the es scroll, we have to send the scroll id in the body, i applied below change and my scroll is maintained at the max limit i set.
self.thread_data.current_es.clear_scroll(body={'scroll_id': scroll_id})

Issue #2305, opened 2019-06-20
Iso Timestamp Type Truncating Milliseconds


JeffAshton:
It appears that the milliseconds are getting truncated for `timestamp_type: "iso"`.  This results in a bad request when the corresponding elasticsearch template uses the `strict_date_time` format for the date type.

The majority of the time this is obviously not an issue, but on rare occasions I'm seeing the following errors:

**Example 1:**

```
{
  "range": {
    "time": {
      "gt": "2019-06-20T13:34:48.688727Z",
      "lte": "2019-06-20T13:34:49Z"
    }
  }
}
```

Error running count query: RequestError(400, u'search_phase_execution_exception', u'failed to parse date field [2019-06-20T13:34:49Z] with format [strict_date_time]')

**Example 2:**

```
{
  "range": {
    "time": {
      "gt": "2019-06-19T12:41:59.981281Z",
      "lte": "2019-06-19T12:42:00Z"
    }
  }
}
```

Error running count query: RequestError(400, u'search_phase_execution_exception', u'failed to parse date field [2019-06-19T12:42:00Z] with format [strict_date_time]')

**Example 3:**

```
{
  "range": {
    "time": {
      "gt": "2019-06-19T02:00:48.783662Z",
      "lte": "2019-06-19T02:00:49Z"
    }
  }
}
```

Error running count query: RequestError(400, u'search_phase_execution_exception', u'failed to parse date field [2019-06-19T02:00:49Z] with format [strict_date_time]')

JeffAshton:
The issue lies here:

https://github.com/Yelp/elastalert/blob/09a3eb609972e6cfec1461ea28de1568a0528734/elastalert/util.py#L141

And how `isoformat` behaves.   From https://docs.python.org/2/library/datetime.html#datetime.datetime.isoformat):

Return a string representing the date and time in ISO 8601 format, YYYY-MM-DDTHH:MM:SS.mmmmmm or, if microsecond is 0, YYYY-MM-DDTHH:MM:SS

JeffAshton:
Hmm, actually it looks like the milliseconds are being truncated here:

https://github.com/Yelp/elastalert/blob/09a3eb609972e6cfec1461ea28de1568a0528734/elastalert/util.py#L148

Qmando:
Hm, it's been too long since I wrote that, I can't remember if there was a practical purpose or not. I know that we've never used `strict_date_time` which is why it was ignored.

[...]

PR #2309, opened 2019-06-24
Adding iso_us timestamp type which guarantees microseconds are not truncated

JeffAshton:
Solution for issue https://github.com/Yelp/elastalert/issues/2305

### DIFF
--- a/elastalert/config.py
+++ b/elastalert/config.py
@@ -16,6 +16,7 @@ import yaml.scanner
 from envparse import Env
 from opsgenie import OpsGenieAlerter
 from staticconf.loader import yaml_loader
+from util import dt_to_iso_microseconds
 from util import dt_to_ts
 from util import dt_to_ts_with_format
 from util import dt_to_unix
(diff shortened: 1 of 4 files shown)


elastalert/config.py
```diff
@@ -235,6 +236,9 @@ def load_options(rule, conf, filename, args=None):
     if rule['timestamp_type'] == 'iso':
         rule['ts_to_dt'] = ts_to_dt
         rule['dt_to_ts'] = dt_to_ts
+    elif rule['timestamp_type'] == 'iso_us':
```
JeffAshton: us => microsecond

Happy to rename to the full name of clarity.

elastalert/util.py
```diff
@@ -148,6 +148,21 @@ def dt_to_ts(dt):
     return ts.replace('000+00:00', 'Z').replace('+00:00', 'Z')
 
 
+def dt_to_iso_microseconds(dt):
+    ts = dt.isoformat()
```
JeffAshton: https://docs.python.org/2/library/datetime.html#datetime.datetime.isoformat

PR #2333, opened 2019-07-12
Fixed top_count_keys not showing in message for compound query_key and Added feature for OpsGenie integration

SabobaS:
When user chooses a **compound query key**, the split is not performed correctly due to a blank in text.

Therefore, a malformed query is performed to Elasticsearch.

This reflects in results not available when using **top_count_keys** directive in rule file.


SabobaS:
Added support to use match fields to populate "opsgenie_alias" in integration with **OpsGenie** (this was already available for "opsgenie_alias").

PR #2335, opened 2019-07-13
Fix Writeback Index Prefix in Example Config

abhishekjiitr:
Change writeback index prefix in `config.yaml.example` from `elastalert_status` to `elastalert` since we already attach appropriate suffixes in `create_index.py`


nsano-rururu:

Issue #2274, opened 2019-06-03
Error executing elastalert-test-rule during args/conf readout


malud:
I am trying to get the latest elastalert version to run with ELK 6 Stack.. however:

executing `elastalert-test-rule --config config.yaml my-rule.yaml` throws:

```
...
Traceback (most recent call last):
File "/usr/bin/elastalert-test-rule", line 11, in <module>
     load_entry_point('elastalert==0.2.0b2', 'console_scripts', 'elastalert-test-rule')()
   File "/usr/lib/python2.7/site-packages/elastalert/test_rule.py", line 441, in main
     test_instance.run_rule_test()
   File "/usr/lib/python2.7/site-packages/elastalert/test_rule.py", line 408, in run_rule_test
     conf = load_conf(args, defaults, overwrites)
   File "/usr/lib/python2.7/site-packages/elastalert/config.py", line 47, in load_conf
     configure_logging(args, conf)
   File "/usr/lib/python2.7/site-packages/elastalert/config.py", line 106, in configure_logging
     if args.verbose and args.debug:
AttributeError: 'Namespace' object has no attribute 'verbose'
```

config content:

```yaml
rules_folder: /opt/elastalert/rules/

buffer_time:
  minutes: 30

run_every:
  seconds: 10

es_host: elasticsearch-status
es_port: 9200

writeback_index: elastalert_status
```

I have tried all suggestions mentioned in #2204

Atem18:
Same issue here with an older ELK. And from what i see, it has nothing to do with the version of ELK.

zffocussss:
> Same issue here with an older ELK. And from what i see, it has nothing to do with the version of ELK. The error is gone if you run Elastalert with --verbose or --debug.

really?

malud:
> Same issue here with an older ELK. And from what i see, it has nothing to do with the version of ELK. The error is gone if you run Elastalert with --verbose or --debug.

I am talking about `elastalert-test-rule` which has no verbose or debug option.

Error thrown: `elastalert-test-rule: error: unrecognized arguments: --verbose`

```
usage: elastalert-test-rule [-h] [--schema-only] [--days DAYS] [--start START]
                            [--end END] [--stop-error] [--formatted-output]
                            [--data FILENAME] [--alert] [--save-json FILENAME]
                            [--use-downloaded]
                            [--max-query-size MAX_QUERY_SIZE] [--count-only]
                            [--config CONFIG]
                            rule
```

[...]

PR #2339, opened 2019-07-16
Fix test rule in beta

### LINKED ISSUE #2303: AttributeError: 'Namespace' object has no attribute 'verbose'
I am facing the issue when i run elastalert-test-rule
what can be issues

i am using 
elasticsearch 7.x

installed elastalert using pip 
elastalert (0.2.0b2)
elasticsearch (7.0.2)

Qmando:
Fixes #2303, #2274

In #2030, a bug was introduced where the argument file passed to load_conf doesn't contain elastalert's required arguments, causing the above issue. This fixes it by setting some defaults.

I don't think these values end up actually being used because we instantiate the ElastAlerter object later and it parses it's own args. This one is just used by the rule loader.

### DIFF
--- a/elastalert/test_rule.py
+++ b/elastalert/test_rule.py
@@ -402,6 +402,13 @@ class MockElastAlerter(object):
         overwrites = {
             'rules_loader': 'file',
         }
+
+        # Set arguments that ElastAlerter needs
+        args.verbose = args.alert
+        args.debug = not args.verbose
+        args.es_debug = False
+        args.es_debug_trace = False
+
         conf = load_conf(args, defaults, overwrites)
         rule_yaml = conf['rules_loader'].get_yaml(args.file)
         conf['rules_loader'].load_options(rule_yaml, conf, args.file)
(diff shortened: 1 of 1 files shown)


elastalert/test_rule.py
```diff
@@ -402,6 +402,13 @@ def run_rule_test(self):
         overwrites = {
             'rules_loader': 'file',
         }
+
+        # Set arguments that ElastAlerter needs
+        args.verbose = args.alert
+        args.debug = not args.verbose
```
r24mille: I'm just a random person from the internet, not a project maintainer, but how do you feel about the following?

```suggestion
        args.debug = not args.alert
```

I realize it's functionally equivalent to your implementation, but feels a bit more consistent with line 232, https://github.com/Yelp/elastalert/blob/ecf40ad31530c34c522d2f8162f6df8f4b32fb1c/elastalert/test_rule.py#L232

Also the conditional on lines 304-307, https://github.com/Yelp/elastalert/blob/ecf40ad31530c34c522d2f8162f6df8f4b32fb1c/elastalert/test_rule.py#L304-L307

It also relates more directly to the help text, 
https://github.com/Yelp/elastalert/blob/ecf40ad31530c34c522d2f8162f6df8f4b32fb1c/elastalert/test_rule.py#L359

iburahim786:
How to get this fix in latest build, i tried in latest docker images, still issue persist.

PR #2358, opened 2019-07-24
Adds to the elasticsearch mapping a dynaic_templates configuration

RubieV:
Now aggregations can be performed on alerts, for example in visualizations. This also enables ElastAlert to run over ElastAlert alerts.


RubieV:
Please note that I have only added this for ES6, an not to ES5. This can be a further contribution.

PR #2371, opened 2019-07-29
Add "http_post_headers" to documentation

watsonlu:


docs/source/ruletypes.rst
```diff
@@ -2051,6 +2053,8 @@ Example usage::
       ip: clientip
     http_post_static_payload:
       apikey: abc123
+    http_post_headers:
+      authorization: Basic 123dr3234
```
r24mille: A basic authorization example seems like a pretty loaded example for this property. However, it does clearly illustrate its utility beyond `foo: bar`.

PR #2372, opened 2019-07-29
Add retries and caching to ES info

Qmando:
- Add some retries to this so that a connectivity blip doesn't bring everything down.
- Cache the elasticsearch_client objects, otherwise we end up seriously spamming the Elasticsearch server with info requests. I think the one place where I pop on es_clients is the only place where rules could change their ES settings.
- Updated a rule to match the new es client caching behavior

### DIFF
--- a/elastalert/__init__.py
+++ b/elastalert/__init__.py
@@ -49,6 +49,8 @@ class ElasticSearchClient(Elasticsearch):
                     self._es_version = self.info()['version']['number']
                     break
                 except TransportError:
+                    if retry == 2:
+                        raise
                     time.sleep(3)
         return self._es_version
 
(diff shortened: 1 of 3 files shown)


elastalert/__init__.py
```diff
@@ -42,7 +44,12 @@ def es_version(self):
         Returns the reported version from the Elasticsearch server.
         """
         if self._es_version is None:
-            self._es_version = self.info()['version']['number']
+            for retry in range(3):
+                try:
+                    self._es_version = self.info()['version']['number']
+                    break
+                except TransportError:
+                    time.sleep(3)
```
danielpops: if it fails 3 times in a row, you'll leave the version as None, is that going to cause issues downstream? should you raise instead when the retries are exhausted?

Qmando: Oops, good call. Third one should just re-raise. It would have caused a "Nonetype has no method split" type error I believe.

PR #2374, opened 2019-07-30
Add Support for Twilio Copilot

cdmastercom:
Per [Twilio Docs](https://www.twilio.com/docs/sms/services#send-a-message-with-copilot) add support for alerts using Copilot. Compability with the existing Twilio SMS alerting method is maintained.

- Make 'twilio_from_number' optional
- Add setting 'twilio_message_service_sid'
- Add setting 'twilio_use_copilot', defaults to False

In case 'twilio_use_copilot' is False, 'twilio_from_number' is required. Otherwise, 'twilio_message_service_sid' is required.

I've tested this using an Elasticsearch 7.1.1 cluster, both SMS and Copilot alerting work.

### DIFF
--- a/elastalert/alerts.py
+++ b/elastalert/alerts.py
@@ -1536,10 +1536,20 @@ class TwilioAlerter(Alerter):
         client = TwilioClient(self.twilio_account_sid, self.twilio_auth_token)
 
         try:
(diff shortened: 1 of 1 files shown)


nsano-rururu:
@cdmastercom

PR #2435, opened 2019-08-31
Adds two option for more efficient reporting

sherifabdlnaby:
These two options work together to create efficient reports that use count query and top keys along with aggregation.

For example, an hourly report that uses a frequency aggregated rule, which uses count query (but we still need data for the report, so we uses top count query too)

`top_count_absolute_timeframe` will make top count query spans time (alert_time - timeframe)
`aggregation_max_alerts` = 1, will allow us to send only one alert with the top count keys of the entire report.

### DIFF
--- a/elastalert/elastalert.py
+++ b/elastalert/elastalert.py
@@ -530,7 +530,7 @@ class ElastAlerter(object):
             buckets = res['aggregations']['filtered']['counts']['buckets']
         else:
             buckets = res['aggregations']['counts']['buckets']
-        if getattr(self.thread_data,"num_hits",None):
+        if getattr(self.thread_data, "num_hits", None):
             self.thread_data.num_hits += len(buckets)
         lt = rule.get('use_local_time')
         elastalert_logger.info(
--- a/elastalert/elastalert.py
+++ b/elastalert/elastalert.py
@@ -1458,13 +1458,13 @@ class ElastAlerter(object):
         if alert_time is None:
             alert_time = ts_now()
 
-        # Calculate Total Hits (useful in-case of aggregation)
-        total_agg_hits = 0
-        total_agg_matches = len(matches)
-        for match in matches:
-            total_agg_hits += match['num_hits']
-
         if rule.get('aggregation'):
+            # Calculate Total Hits (useful in-case of aggregation)
+            total_agg_hits = 0
+            total_agg_matches = len(matches)
+            for match in matches:
+                total_agg_hits += match['num_hits']
+
             num_matches = rule.get('aggregation_max_alerts', len(matches))
             matches = matches[:num_matches]
             for match in matches:
--- a/elastalert/elastalert.py
+++ b/elastalert/elastalert.py
@@ -1458,6 +1458,10 @@ class ElastAlerter(object):
         if alert_time is None:
             alert_time = ts_now()
 
+        if rule.get('aggregation'):
+            num_matches = rule.get('aggregation_max_alerts', len(matches))
+            matches = matches[:num_matches]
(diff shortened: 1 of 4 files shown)


elastalert/elastalert.py
```diff
@@ -1457,6 +1458,20 @@ def send_alert(self, matches, rule, alert_time=None, retried=False):
         if alert_time is None:
             alert_time = ts_now()
 
+        # Calculate Total Hits (useful in-case of aggregation)
+        total_agg_hits = 0
+        total_agg_matches = 0
+        for match in matches:
+            total_agg_hits += match['num_hits']
+            total_agg_matches += match['num_matches']
```
Qmando: I think this is going to produce a numbers that's way too big. If a single query generates 5 matches, each one is going to have num_matches=5, so the total_agg_matches would be 25. Shouldn't it just be the original length of matches before you do `matches[:num_matches]`?

sherifabdlnaby: But in case of using `use_terms_query: true` where *matches* are per bucket, then `num_matches` -in this for loop's match- equals the number of buckets matched.

We can explicitly document that `total_agg_matches` equals the number of alerts triggered(whatever how many buckets this rule matched).

But my intuition with to keep `num_matches` definition consistent, which is *per bucket* in-case `use_terms_query: true` or `query_key` is defined.

What do you think?

Qmando: When you say bucket, do you really mean aggregation bucket? For a single query using `use_terms_query`, the num_matches are going to all be the same. Look at the lines [here](https://github.com/Yelp/elastalert/blob/0172a8a8d146d5d7a6d0255c5a0cf4d02344682a/elastalert/elastalert.py#L911), those matches could be from different buckets.

For an aggregation, however, each alert could have matches from different queries, so the num_matches are all going to be different. I'm still not entirely sure what the value is for having the total num_matches.

Can you give me a more concrete example to help me understand?
For example if you had a rule like

```
type: frequency
num_events: 1000
timeframe:
  minutes 10
query_key: hostname
use_terms_query: true
aggregation:
  hours: 1
```

If one query got 1001 hits for hostnames A, B, and C, each match would show `num_matches: 3`. Each would be added separately to the aggregation.  Assuming thats the only query that had matches, with this code, it would add them all together and get 9.

Am I misunderstanding? If you are truncating the number of matches, why not just put the original length in?

sherifabdlnaby: First of all, apologies for the late reply, I wasn't able to address them but now.

> I'm still not entirely sure what the value is for having the total num_matches.

This is used with `aggregation_max_alerts`, we were trying to add hourly reports, where a slack message will report last hour's errors/triggers.

Using `aggregation` to do so, but then, sometimes we'd have thousands of alerts, making Slack *very* spammy, as by default it prints a message for every alert triggered during aggregation ( separated with --------- ), and then I introduced the `aggregation_max_alerts` option to avoid printing too many errors.

Actually, for our use case, we set `aggregation_max_alerts: 1` and only printed the top_events count using `use_count_query: true`. so, only left was `total_agg_hits` and `total_agg_matches` to be available on every alert, in-case you limited the `aggregation_max_alerts` but still want to know how many matches/hits occurred during the entire aggregation.

sherifabdlnaby: But I see, looks like my assumption of `num_matches` was wrong. and I see your points, that we're going to have a bigger `num_matches` than there should be.

I'll update the PR.

sherifabdlnaby:
@Qmando I am unable to detect why build fails. if you can help me with that.

PR #2467, opened 2019-09-18
Handling non-compound query keys defined as an array

JeffAshton:
The following `query_key` configurations should be equivalent:

```query_key: status```
```query_key: [ status ]```

I accidentally defined my singular query key as an array, and noticed that the type was forced to a string.  This is because `elastalert.py` assumes `compound_query_key` to always have more than 1 value.  If there is only one element in the array, the value gets overriden with a string.

https://github.com/Yelp/elastalert/blob/a5322f25b80828cc2183970c9c9f81f32867ce3b/elastalert/elastalert.py#L336

### DIFF
--- a/elastalert/loaders.py
+++ b/elastalert/loaders.py
@@ -324,9 +324,15 @@ class RulesLoader(object):
         if 'include' in rule and type(rule['include']) != list:
             raise EAException('include option must be a list')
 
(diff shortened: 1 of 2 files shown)


Qmando:
Thanks for all the PRs!!

PR #2478, opened 2019-09-20
Upgrading jsonschema library to 3.0.2

JeffAshton:


elastalert/schema.yaml
```diff
@@ -1,4 +1,4 @@
-$schema: http://json-schema.org/draft-04/schema#
+$schema: http://json-schema.org/draft-07/schema#
```
JeffAshton: ```draft-07``` has `if` statements which I would like to leverage.

PR #2483, opened 2019-09-23
Bug: import within import not getting imported

aksandal:
Fix: updates the rule_name for every while loop causing it to import as many import within import within


Qmando:
What is the bug here? You can do recursive imports.

Updating the file name is going to have other weird consequences, I'm not sure why we'd want to update it to the imported file.

PR #2484, opened 2019-09-23
Stop job when it gets disabled

Qmando:
As a consequence of moving rules to run on a scheduler, they didn't actually get disabled properly when you removed the file.

Instead of just modifying self.rules, we need to call scheduler.remove_job too. To do that, I had to change the code so that we could grab it's name first before we delete it.

Tested this using a trivial example (some lines cut out)
```
INFO:elastalert:Ran test-rule-2 from 2019-09-23 14:05 PDT to 2019-09-23 14:50 PDT: 0 query hits (0 already seen), 0 matches, 0 alerts sent
INFO:elastalert:Ran test-rule from 2019-09-23 14:05 PDT to 2019-09-23 14:50 PDT: 0 query hits (0 already seen), 0 matches, 0 alerts sent
INFO:elastalert:Background alerts thread 0 pending alerts sent at 2019-09-23 14:51 PDT
INFO:elastalert:Rule file elastalert_rules/test_2.yaml not found, stopping rule execution
INFO:elastalert:Background configuration change check run at 2019-09-23 14:51 PDT
INFO:elastalert:Disabled rules are: []
INFO:elastalert:Sleeping for 29.999797 seconds
INFO:elastalert:Ran test-rule from 2019-09-23 14:06 PDT to 2019-09-23 14:51 PDT: 0 query hits (0 already seen), 0 matches, 0 alerts sent
INFO:elastalert:Background alerts thread 0 pending alerts sent at 2019-09-23 14:51 PDT
INFO:elastalert:Background configuration change check run at 2019-09-23 14:51 PDT
INFO:elastalert:Disabled rules are: []
INFO:elastalert:Sleeping for 29.999795 seconds
...
```

### DIFF
--- a/elastalert/elastalert.py
+++ b/elastalert/elastalert.py
@@ -23,8 +23,8 @@ from apscheduler.schedulers.background import BackgroundScheduler
 from croniter import croniter
 from elasticsearch.exceptions import ConnectionError
 from elasticsearch.exceptions import ElasticsearchException
-from elasticsearch.exceptions import TransportError
 from elasticsearch.exceptions import NotFoundError
+from elasticsearch.exceptions import TransportError
 
 from . import kibana
 from .alerts import DebugAlerter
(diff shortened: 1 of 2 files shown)


elastalert/elastalert.py
```diff
@@ -1067,7 +1067,13 @@ def load_rule_changes(self):
             if rule_file not in new_rule_hashes:
                 # Rule file was deleted
                 elastalert_logger.info('Rule file %s not found, stopping rule execution' % (rule_file))
-                self.rules = [rule for rule in self.rules if rule['rule_file'] != rule_file]
+                for rule in self.rules:
+                    if rule['rule_file'] == rule_file:
+                        break
```
danielpops: what if there is more than one disabled rule? won't this only disable one of them?

Qmando: See line 1066. This is in a loop over all rule hashes

PR #2493, opened 2019-09-30
Fixing slack ssl verification

JeffAshton:


elastalert/alerts.py
```diff
@@ -1209,7 +1209,7 @@ def alert(self, matches):
                     if self.slack_ca_certs:
                         verify = self.slack_ca_certs
                     else:
-                        verify = self.slack_ignore_ssl_errors
+                        verify = not self.slack_ignore_ssl_errors
```
JeffAshton: If ```slack_ignore_ssl_errors = True```, then you don't want to validate.

JeffAshton: Rewriting it as such might make it more clear and explicit:

```
                    if self.slack_ca_certs:
                        verify = self.slack_ca_certs
                    elif self.slack_ignore_ssl_errors:
                        verify = False
                        requests.packages.urllib3.disable_warnings()
                    else:
                        verify = True
```

JeffAshton: https://requests.readthedocs.io/en/master/user/advanced/#ssl-cert-verification

JeffAshton:
Ping @Qmando

PR #2495, opened 2019-10-02
Adding a timeout to all requests calls

ccottam:
There  is no default timeout value for the python requests module - some alerts had a default, most did not, I have duplicated the format used by the http post alert to all the other alerts using a requests post.

All have a default of 10 seconds, overridable by a config item


ccottam:
I believe broken tests are related to the release of elasticsearch 7.0.5

Issue #2036, opened 2018-12-18
Relaert is not working in frequency rule type


skatakar:
I am using below parameters in frequency rule type but realert is not working on specified time duration. I have mentioned relaert with hours : 1 and queary_key : "Hostname" ie since realert is 1 hour I assume it should aggregate records based on hostname field for the hour (assume from 9:00 to 10:00) and as num_events : 1 hence it should raise only one alarm per hostname eventhough hostname have more events in that hr. it is working as expected. But if the event comes after one hr ie at 10:25, the alarm is not getting generated. currently I haven't configured any alert notification. I can see them are supressed in console.

rule file has:
es_host: localhost
es_port: 9200
rules_folder : "D:\elastalert-master\example_rules"
buffer_time :
hours : 4

timeframe:
days : 4

writeback_index : test_elastalert1
name: test5 rule
type: frequency
timestamp_field: "time"
index: test_elastalert1
num_events: 1
aggregation:
#minutes: 10
days : 1
aggregation_key : "Hostname"
query_key : "Hostname"

realert :
hours : 1
filter:

query:
match :
"event_message" : "Ping"
alert: "email"
alert_subject: "Issue is occurred "
alert_info: "event from "

elastalert-test-rule

D:\elastalert-master\example_rules>elastalert-test-rule test5.yaml
Successfully loaded test5 rule

Got 4 hits from the last 1 day

Available terms in first hit:
event_message
event_sev
Hostname
event_source
time

INFO:elastalert:Note: In debug mode, alerts will be logged to console but NOT actually sent.
To send them but remain verbose, use --verbose instead.
INFO:elastalert:Queried rule test5 rule from 2018-12-16 14:13 Central Standard Time to 2018-12-16 18:13 Central Standard Time: 0 / 0 hits
INFO:elastalert:Queried rule test5 rule from 2018-12-16 18:13 Central Standard Time to 2018-12-16 22:13 Central Standard Time: 0 / 0 hits
INFO:elastalert:Queried rule test5 rule from 2018-12-16 22:13 Central Standard Time to 2018-12-17 02:13 Central Standard Time: 0 / 0 hits
INFO:elastalert:Queried rule test5 rule from 2018-12-17 02:13 Central Standard Time to 2018-12-17 06:13 Central Standard Time: 0 / 0 hits
[...]

Issue #1047, opened 2017-04-24
Incorrect timestamp


kernelpanicfbsd:
Hello all, I'm using elastalert to query a winlogbeat index which does use @timestamp for the date/time.

However, the email alert I'm receiving has a timestamp value that is an hour behind the actual timestamp on the document it has matched; for example a document with the timestamp of @timestamp: 2017-04-24T**10:05**:42.687Z in the elastalert database has a timestamp of @timestamp: 2017-04-24T**09:05**:42.687Z on the elastalert email.

Can anyone help?

Qmando:
Are you sure you're not looking in Kibana, where the timestamp is automatically adjusted to local time? Can you post the raw JSON of the document that matched (including _id) along with the alert (including _id)?

kernelpanicfbsd:
Hello thanks for getting back to me -  yes, I was looking in Kibana, however my understanding is that by default Elastalert uses local time too, so shouldn't the Kibana and Elastalert timestamps be the same value?

Qmando:
Internally ElastAlert converts everything into UTC (The Z at the end of the timestamps you posted means UTC in iso8601). In a few places it will convert it into local time to display in the alert, but when it does it will add the timezone.

kernelpanicfbsd:
OK thanks, so what should I do to get the two timestamp values to agree with each other?

Qmando:
You can set ``use_local_time: true`` in your config. This will at least make the log lines and title of the alert have the local timezone. In Logstash you can add a second timestamp field, converted into local time, and then put that field into your alert_text

kernelpanicfbsd:
Thanks again - is use_local_time: true set on config.yaml or in a rule?

Qmando:
config.yaml to make it apply to all rules, or per rul

kernelpanicfbsd:
Adding the use_local_time option has not made a difference, the alert @timestamp is still an hour behind so I assume I'll need to go the Logstash route?

Test ElastAlert Rule

@timestamp: 2017-04-24T**18:24**:21.841Z
_id: AVuhNQq9ZFqcPo9Hyrv8
_index: winlogbeat-2017.04
_type: wineventlog
event_data: {
    "MemberName": "####################", 
    "SubjectUserName": "############", 
    "TargetUserName": "######"
}
event_id: 4728
message: A member was added to a security-enabled global group.

Subject:
	Security ID:		#####################
	Account Name:		#############
	Account Domain:		#############
[...]

PR #2542, opened 2019-10-29
Enable to select whether to prohibit enhancement from being run on alert information to be written back to ES

perceptron01:
# Background and problem

With running enhancements, the data type of elements in a match could be changed from original one due to the enhancements.
As a result, when the data type that Elasticsearch's writeback_index expects does not match the data type in a match to be written to ES, trying to write alert info to Elasticsearch will fail. 

This problem especially tends to occur when using enhancements that change the format of @timestamp (or timestamp_field).
In that case, the following error would appear.

```
ERROR:root:Error writing alert info to Elasticsearch: RequestError(400, 'mapper_parsing_exception', "failed to parse field [match_time] of type [date] in document with id 'XXXXXXXXXXX'")
Traceback (most recent call last):
  File "/<path_to>/venv/lib/python3.6/site-packages/elastalert/elastalert.py", line 1625, in writeback
    res = self.writeback_es.index(index=index, body=body)
  File "/<path_to>/venv/lib/python3.6/site-packages/elasticsearch/client/utils.py", line 84, in _wrapped
    return func(*args, params=params, **kwargs)
  File "/<path_to>/venv/lib/python3.6/site-packages/elasticsearch/client/__init__.py", line 364, in index
    "POST", _make_path(index, doc_type, id), params=params, body=body
  File "/<path_to>/venv/lib/python3.6/site-packages/elasticsearch/transport.py", line 350, in perform_request
    timeout=timeout,
  File "/<path_to>/venv/lib/python3.6/site-packages/elasticsearch/connection/http_requests.py", line 156, in perform_request
    self._raise_error(response.status_code, raw_data)
  File "/<path_to>/venv/lib/python3.6/site-packages/elasticsearch/connection/base.py", line 181, in _raise_error
    status_code, error_message, additional_info
elasticsearch.exceptions.RequestError: RequestError(400, 'mapper_parsing_exception', "failed to parse field [match_time] of type [date] in document with id 'XXXXXXXXXXX'")
```

Several related issues have also been reported.
#1047 #2036

# Cause
The cause of this problem is that, in the current code, it is the alert info modified by enhancements that will be written back to ES. 

# Solution
Add an option which realizes that enhancements will be run only on the content of alerts send by Alerters, whereas info about the alerts to be written back into Elasticsearch will not processed by enhancements.

### DIFF
--- a/docs/source/ruletypes.rst
+++ b/docs/source/ruletypes.rst
@@ -625,6 +625,11 @@ run_enhancements_first
 or dropped before affecting realert or being added to an aggregation. Silence stashes will still be created before the
 enhancement runs, meaning even if a ``DropMatchException`` is raised, the rule will still be silenced. (Optional, boolean, default false)
 
+block_enhancements_in_writeback
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+``block_enhancements_in_writeback``: If true, enhancements will be run only on the content of alerts send by Alerters,
+whereas info about the alerts to be written back into Elasticsearch will not processed by enhancements. (Optional, boolean, default false)
+
 query_key
 ^^^^^^^^^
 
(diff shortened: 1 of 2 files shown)


nsano-rururu:
Test are failing

PR #2570, opened 2019-11-18
updated rules_mapping to include spike_aggregation

pgardiner:
Was getting message "elastalert.util.EAException: Error loading file rules/apache_busy.yaml: Could not import module spike_aggregation: not enough values to unpack (expected 2, got 1)". Adding rule to mapping fixed issue.


mareksimunek:
I know you are probably busy, but could you look into this fix :
@pdscopes @Qmando @danielpops

PR #2572, opened 2019-11-19
run on single writeback index option

hashtag-24:
Hello there,
I have been working on this for some time.
We are using index as a service and therefore needed to simplify things so that elastalert would be able to run on a single index.
I tested this on elasticsearch 6.8 (including create_index) but do not have the means to test on 5.x or 7.x


hashtag-24:
something is missing, working on it

Issue #2580, opened 2019-11-25
top_count_keys per query_key


dream91:
I use top_count_keys and I noticed that it is not counted per query_key. Is it somehow possible to count per query_key?

Qmando:
This is supported, but it's very possible there's a bug. Are you using `min_cardinality` or is your query_key a nested field? I think those might cause this problem.

dream91:
```
[..]
top_count_keys: ["username"]
cardinality_field: "username"
max_cardinality: 7
query_key: event_data.IpAddress
[..]
```

Yes, query_key is the nested field.

Qmando:
You could try this change:
```
--- a/elastalert/elastalert.py
+++ b/elastalert/elastalert.py
@@ -1467,10 +1467,7 @@ class ElastAlerter(object):
         # Compute top count keys
         if rule.get('top_count_keys'):
             for match in matches:
-                if 'query_key' in rule and rule['query_key'] in match:
-                    qk = match[rule['query_key']]
-                else:
-                    qk = None
+                qk = lookup_es_key(match, rule['query_key'])
```

I'll try to get this fixed in master

PR #2593, opened 2019-12-05
Allow run_every to be unique per rule

### LINKED ISSUE #497: How to specify different run_every: values on rules
I have some rules I want to run every 10 seconds, and some that can run only once a day.  But I can't seem to find a way to override the run_every: value in the base config file.  Is this currently supported?

Qmando:
Fixes #497 

It's currently copying run_every as part of the "default rule properties", which used to be a set of things that are not configurable by the rule yaml itself. This was never removed despite run_every becoming a per rule property.

Several tests were relying on this because they are mocking out the config loader, which also adds a default run_every, and not specifying one.

I added a test for specifically this feature, and confirmed that without the changes to elastalert.py, that new test fails.

As a bonus, I've included a fix for #2580, because I haven't created a separate PR for it yet. This allows nested query keys to work in top_count_keys.

Also note: I also added a pin for elasticsearch==7.0.0, because apparently 7.1.0 will NOT work with ES < 6.6 due to it not supported _source_include(s?). 7.0.0 does. Tests won't pass otherwise.

### DIFF
--- a/elastalert/elastalert.py
+++ b/elastalert/elastalert.py
@@ -27,10 +27,10 @@ from elasticsearch.exceptions import NotFoundError
 from elasticsearch.exceptions import TransportError
 
 from . import kibana
-from .kibana_discover import generate_kibana_discover_url
 from .alerts import DebugAlerter
 from .config import load_conf
 from .enhancements import DropMatchException
+from .kibana_discover import generate_kibana_discover_url
 from .ruletypes import FlatlineRule
 from .util import add_raw_postfix
 from .util import cronite_datetime_to_timestamp
@@ -1022,8 +1022,7 @@ class ElastAlerter(object):
                            'processed_hits',
                            'starttime',
                            'minimum_starttime',
-                           'has_run_once',
-                           'run_every']
+                           'has_run_once']
         for prop in copy_properties:
             if prop not in rule:
                 continue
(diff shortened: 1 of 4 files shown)


yuriipolishchuk-finsight:
@Qmando is there any blocker for this PR to be merged?

yuriipolishchuk-finsight:
@Qmando thank you, sir!  🎉

moix:
Hi, does the restriction of elasticsearch==7.0.0 means we cannot run against any other el version? I have a cluster in 7.5.1 so initially was forcing the elasticsearch library to 7.5.1 but after upgrading elastalert to 0.2.4 it fails.
Thanks!

Qmando:
What does "it fails" mean exactly

moix:
requirements are set to force elasticsearch==7.0.0 so it complains if I have 7.5.1

Qmando:
Try `$ python -m elastalert.elastalert` instead of `$ elastalert`

rashmichoudhary07:
@Qmando 
So is allow run_every to be unique per rule working now? i tried it at my end, but it still considers the run_every from config.yaml only for all my rules.

Please help!
Thanks

MakoWish:
+1 looking for this feature to work.

PR #2596, opened 2019-12-07
Add a new rule: find_match

karlos1841:
Hi guys!

I'd like you to take a look at the rule I wrote some time ago for a customer.
For option invert: false the rule is an equivalent of elapsed filter plugin for logstash meaning an event gets marked as start event when value of compare_key field is equal to start_value.
An event gets marked as end event when value of compare_key field is equal to end_value.
Elapsed time between those two events with the same query_key is calculated.
If the calculated time is within specified timeframe we have an alert.

For invert: true we are alerting when end event is not present within specified timeframe.

Sample rule:

name: test
type: find_match
index: test*
query_key: id
compare_key: message
start_value: "start"
end_value: "end"
invert: false
timeframe:
  minutes: 1

alert:
  - command
command: ['/bin/echo', 'ALERT']

### DIFF
--- a/elastalert/loaders.py
+++ b/elastalert/loaders.py
@@ -50,6 +50,7 @@ class RulesLoader(object):
         'cardinality': ruletypes.CardinalityRule,
         'metric_aggregation': ruletypes.MetricAggregationRule,
         'percentage_match': ruletypes.PercentageMatchRule,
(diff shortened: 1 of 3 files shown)


nsano-rururu:
Test are failing

nsano-rururu:
@karlos1841

The branch is "alt". Please be careful.

PR #2604, opened 2019-12-10
Task/show warn false

alvarolmedo:
Using the ssl_show_warn option of elasticsearch's library from elastalert's config file.

This options reduces the warnings when a ssl connection is performed and it's no verified.


alvarolmedo:
The CI seems broken with old ES's versions.

alvarolmedo:
@Qmando this option is useful for us because our certs are selfsigned and they aren't verified.

nsano-rururu:
Test are failing

PR #2605, opened 2019-12-10
Task/zabbix fixes

alvarolmedo:
Some problems with timestamp and no microseconds provided solved.


alvarolmedo:
Included in #2640

PR #2647, opened 2020-01-21
Add option to generate and attach Kibana link to Jira

keuko:


keuko:
It seems tests are broken , could someone fix it ?

PR #2658, opened 2020-01-27
Jira attach kibana link

keuko:
This patch is adding feature to add kibana link to jira task.


nsano-rururu:
@keuko

PR #2681, opened 2020-02-14
Add parameter 'smtp_ca_file', fix STARTTLS problem

daskanu:
Hey guys,

I added a new parameter 'smtp_ca_file' in order to configure a CA file and therefore I added the usage of [ssl.SSLContext](https://docs.python.org/3.4/library/ssl.html#ssl.SSLContext) to the code.

I also added a check for the Python version since context [was added after 3.3](https://docs.python.org/3.4/library/smtplib.html#smtplib.SMTP.starttls).

### DIFF
--- a/elastalert/alerts.py
+++ b/elastalert/alerts.py
@@ -5,6 +5,7 @@ import json
 import logging
 import os
 import re
+import ssl
 import subprocess
 import sys
 import time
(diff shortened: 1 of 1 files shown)


nsano-rururu:
It seems that elastalert/tests/alerts_test.py needs to be modified.
Check the "Detail" of "continuous-integration/travis-ci/pr — The Travis CI build failed".

nsano-rururu:
Here are some things to keep in mind when making a pull request to elastalert2. Please note that the source code structure of the alert is now different from the original yelp / elastalert.

Structure of related files and source code

```
elastalert
  loaders.py
  schema.yaml
  /alerters
    email.py
  /tests
    /alerters
      email_test.py
```

PR #2684, opened 2020-02-18
Pin elasticsearch to 7.0.0 in requirements.txt

makefu:
in accordance with fix_497 , the pin seems to be missing in the `requirements.txt`


nsano-rururu:
Reflected in elastalert2

nsano-rururu (approved):

PR #2702, opened 2020-03-04
Update Docs for query_key types (compound_query_key)

cawalch:
Compound query keys must be passed as a list of strings. This updates the docs to reflect the possible `query_key` types.

Reference where `query_key` type is checked for setting the `compound_query_key`
https://github.com/Yelp/elastalert/blob/2791119dc1e31abe5c46bf1b733d40de90ed37a1/elastalert/loaders.py#L332


nsano-rururu:
Test are failing

Issue #1926, opened 2018-09-24
--patience flag and disable_rules_on_error doesn't seem to work


chrislujan:
version:
```
$ sudo pip list | grep elastalert
elastalert                       0.1.33
```

I have a functioning config but I would like elastalert to be patient as elasticsearch sometimes takes a minute to come up after a restart. I have tried using --patience with no luck. I'm not sure what I'm doing wrong. I have also noticed that if ES is not responding, rules are disabled, regardless of my use of `disable_rules_on_error` (see 5th warning of log). Is it possible that my configurations aren't being applied?

Here's my config:

```
rules_folder: /etc/elastalert/rules

run_every:
  minutes: 1

buffer_time:
  minutes: 15

es_host: 127.0.0.1

es_port: 9201

writeback_index: elastalert

alert_time_limit:
  hours: 3

disable_rules_on_error: false
```

Here's my command and output: 

```
$ time sudo /usr/bin/python2 /usr/bin/elastalert --verbose --patience minutes=5 --config config.yaml
WARNING:elasticsearch:GET http://127.0.0.1:9201/ [status:N/A request:0.002s]
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/elasticsearch/connection/http_requests.py", line 76, in perform_request
    response = self.session.send(prepared_request, **send_kwargs)
  File "/usr/lib/python2.7/site-packages/requests/sessions.py", line 622, in send
    r = adapter.send(request, **kwargs)
  File "/usr/lib/python2.7/site-packages/requests/adapters.py", line 513, in send
    raise ConnectionError(e, request=request)
ConnectionError: HTTPConnectionPool(host='127.0.0.1', port=9201): Max retries exceeded with url: / (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x2a78590>: Failed to establish a new connection: [Errno 111] Connection refused',))
WARNING:elasticsearch:GET http://127.0.0.1:9201/ [status:N/A request:0.003s]
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/elasticsearch/connection/http_requests.py", line 76, in perform_request
    response = self.session.send(prepared_request, **send_kwargs)
  File "/usr/lib/python2.7/site-packages/requests/sessions.py", line 622, in send
    r = adapter.send(request, **kwargs)
  File "/usr/lib/python2.7/site-packages/requests/adapters.py", line 513, in send
    raise ConnectionError(e, request=request)
[...]

PR #2707, opened 2020-03-09
Respect disable_rules_on_error flag for connection errors

rdepril:
Respect `disable_rules_on_error` flag for connection errors during `init_rule`.


rdepril:
I think this solves one of the issues in #1926

PR #2743, opened 2020-04-16
Added back custom fields to the hive alerter

Qmando:
Using config `customFields: {'message': {'type':'string', 'value':'{match[message]}'}, 'foo': {'type': 'string', 'value': 'qlo msg 2'}}` with data in the match `'message': 'qlo msg 1'`

Segment of the post data from the old alerter:
```
"customFields": {
        "foo": {
            "order": 1,
            "string": "qlo msg 2"
        },
        "message": {
            "order": 0,
            "string": "qlo msg 1"
        }
    },
```

And with the new one:
```
   "customFields": {
        "foo": {
            "order": 1,
            "string": "qlo msg 2"
        },
        "message": {
            "order": 0,
            "string": "qlo msg 1"
        }
    },
```

which are identical.

### DIFF
--- a/elastalert/alerts.py
+++ b/elastalert/alerts.py
@@ -2145,9 +2145,15 @@ class HiveAlerter(Alerter):
                 'date': int(time.time()) * 1000
             }
             alert_config.update(self.rule.get('hive_alert_config', {}))
-
+            custom_fields = {}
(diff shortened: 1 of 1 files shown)


golaso:
Hello, is there any hive version constrains to use this feature?
I tried to push alert contains customFields to the hive but see nothing in hive.

PR #2760, opened 2020-04-24
Fixing slack ssl verification

JeffAshton:


elastalert/alerts.py
```diff
@@ -1205,7 +1205,7 @@ def alert(self, matches):
                     if self.slack_ca_certs:
                         verify = self.slack_ca_certs
                     else:
-                        verify = self.slack_ignore_ssl_errors
+                        verify = not self.slack_ignore_ssl_errors
```
JeffAshton, line 1208: If slack_ignore_ssl_errors = True, then you don't want to validate.

https://requests.readthedocs.io/en/master/user/advanced/#ssl-cert-verification

jertel:
According to these (unofficial) Python Requests docs, https://www.w3schools.com/python/ref_requests_post.asp, the verify parameter is indeed expecting a True argument if the intention is to verify certificates. Thanks for the submission.

nsano-rururu:
Merged into elastalert2.
https://github.com/jertel/elastalert2/pull/30

PR #2778, opened 2020-05-02
Sns docs update

nsano-rururu:
Since the documentation on SNS is old,
aws_access_key, aws_secret_key, profile is
You need to update to aws_access_key_id, aws_secret_access_key, and aws_profile.


nsano-rururu:
Since the branch was in a strange state, close it once

PR #2794, opened 2020-05-10
Fix Zabbix(Docs & schema.yaml)

nsano-rururu:
Although it is described as "zbx_item" in the document,
 it is in a state different from the implementation. Correctly "zbx_key"


nsano-rururu:
@Qmando can you review?

nsano-rururu:
@Qmando can you review?

PR #2813, opened 2020-05-17
added docs opsgenie_proxy

nsano-rururu:
Added because there is no description of opsgenie_proxy


nsano-rururu:
@Qmando can you review?

PR #2826, opened 2020-05-26
Removing/Disabling rule should stop scheduler jobs

jonscyr:
when is_enabled is mutated at runtime, alerts still keep coming in. Disabled alerts state is not updated. This should probably fix it.


nsano-rururu:
Test are failing

virtyaluk:
@jonscyr, any news from you? Thanks.

PR #2838, opened 2020-06-04
Fixes to is_enabled state changes during runtime

jasonneurohr:
The is_enabled flag is not correctly handled when changed between true and false state during runtime. This fixes the implementation so it is handled correctly.

This PR also fixes a small issue in the example rules.


nsano-rururu:
Test are failing

PR #2858, opened 2020-06-23
fix attribute error is raised when query ran for future time

Swapnil811:
when time is in future we simply return 0 , attributes - num_hits, num_dupes, cumulative_hits remains uninitialised which are used for logging info about the query and hence attribute error is raised.


nsano-rururu:
Merged into elastalert2.
https://github.com/jertel/elastalert2/pull/44

nsano-rururu (approved):

Issue #2182, opened 2019-03-27
thehive alerter: allow using field values for TLP and severity


problematiq:
If not already possible, can we get the ability to use field values for the `tlp:` and `severity:` for hive alert? e.g:
```
tlp: '{match[alert.tlp]}'
severity: '{match[alert.severity]}'
```

ptobis:
I vote for it and I also suggest to have this ability in the `customFields:` in TheHive alerter.

mdennison-ntc:
I would also like this feature. Is it possible at the moment? What needs to be done to get this working?

PR #2891, opened 2020-07-20
TheHive alerter: Allow severity and tlp to be set by rule

defensivedepth:
Explicitly casts severity/tlp as `int`, which is required by TheHive api.

Closes https://github.com/Yelp/elastalert/issues/2182

@Qmando can you review?


weslambert:
I think this would be awesome to have!  🙏

PR #2896, opened 2020-07-25
Remove Stride integration

nsano-rururu:
The Stride service was shutdown on February 15th, 2019.


nsano-rururu:
@Qmando can you review?

PR #2898, opened 2020-07-29
Fix slack kibana properties in schema

varunvora:
Made the schema valid according to draft-07


varunvora:
@Qmando can you review?

PR #2993, opened 2020-10-21
Add Jinja2 Templating Option to Alert Text Formatting 🌈

sherifabdlnaby:
Hello @Qmando ,

I've been maintaining a fork of ElastAlert that's used internally for many changes, one of the most useful and important change is having **Jinja2 Templating Engine** in Alert Text.

## Motivation

Jinja2 allows alerts to be more **concise**, **informative**, **clean**, and more **dynamic**, it allows us for example:
1. Trim long texts
2. Add Conditional Message based on data.
3. Use For loops to print dynamic data nicely.
4. Calculate the Sum of an array and all other aggregations.
5. Round Decimal Numbers and all sorts of transformations.
6. Ability to have conditions, comparisons, and some scripting in formatting the alert.

> Basically everything in [Jinja2 Filters Docs](https://jinja.palletsprojects.com/en/2.11.x/templates/#list-of-builtin-filters)

**Adding Jinja2 improved Slack alerts tremendously**, we would send a `@channel` or `@here` based on the count of `num_matches` or based on a certain error type. We would also have a multi-nested for-each loop that prints out data nicely in the alert and trimming possible long messages so the alert still is small and concise. Also, email alerts were much nicer and more cleaner.

## Example:
```
alert_text_type: alert_text_jinja
alert_text: |
  An alert in *`{{infra.service}}`* triggered after matching *({{num_hits}} hit!)*
  {% if infra.service == 'customer_service' %} <!here> :mega: Critical Service Error {% endif %}
  > :kibana: <{{kibana_link}}|Check event timeframe in Kibana>
  ---------- *Error Context*
  {% for key, value in context.items() %}  - {{key}}: {{value|truncate(256,True)}} {% endfor %}
  ---------- *Error Message Preview*
  > ```
    {{message|truncate(256,True)}
    ```
  ---------- *Infra Context*
  *Container*: `{{infra.container}}` *Image*: `{{infra.image}}` *Host*: `{{infra.host}}`

```

Generates an alert like:
![Screen Shot 2020-10-21 at 4 20 35 AM](https://user-images.githubusercontent.com/16992394/96673388-ac1b6300-1366-11eb-8c55-cbe085f819c2.png)



Actual Alerts I use internally are much better looking but because I've made edits to fully utilize Slack API. I will try to breakdown one by one and open PRs.


## Changes Made

Only a new `alert_text_type` = `alert_text_jinja` that will render the jinja template in `alert_text` using Jinja2 package.
This change is small and backward compatible it doesn't affect any old behavior only add new.


Thank you.


sherifabdlnaby:
@Qmando Any update about this PR ?

TOoSmOotH:
@sherifabdlnaby Any way you could push this PR to https://github.com/jertel/elastalert ?

Issue #2983, opened 2020-10-15
blist no longer works in Python 3.9


maybe-sybr:
ref: DanielStutzbach/blist#90

`blist` is used for its `sortedlist` in `ruletypes.py` - might I suggest that the `SortedKeyList` [0] from python-sortedcontainers [1] could replace it?

[0] http://www.grantjenks.com/docs/sortedcontainers/introduction.html#sorted-key-list
[1] https://github.com/grantjenks/python-sortedcontainers

nsano-rururu:
It seems that the operation is guaranteed up to 3.6. Therefore, it seems that the operation of 3.9 is not guaranteed.

nsano-rururu:
It's just a personal impression, but I think it's better to change to another library than to continue using blist.
Are you planning to make a pull request to change blist to python-sorted containers?

**File to change**
requirements.txt
setup.py
elastalert/ruletypes.py

nsano-rururu:
ElastAlert shouldn't officially support Python 3.8 as it starts with Syntax Warning from Python 3.8.

**SyntaxWarning: "is" with a literal. Did you mean "=="?**

**_Files that may need to be modified_**

elastalert/
　__init__.py
　alerts.py
　create_index.py
　elastalert.py
　kibana_discover.py
　loaders.py
　opsgenie.py
　ruletypes.py
　test_rule.py
　util.py
tests/
　base_test.py
　elasticsearch_test.py
　kibana_discover_test.py
　loaders_test.py
　rules_test.py
　util_test.py

**SyntaxWarning: "is not" with a literal. Did you mean "!="?**

**_Files that may need to be modified_**

elastalert/
　alerts.py
　config.py
　create_index.py
　elastalert.py
　opsgenie.py
　ruletypes.py
　util.py

nsano-rururu:
If you want to upgrade Python version, you may need to update the following files used by Travis CI

Dockerfile-test 
travis.yml

maybe-sybr:
@nsano-rururu - I can make a PR when i get some time for the replacement of `blist`. It won't be today but perhaps over the weekend.

wrt those syntaxwarnings, I had a quick grep though the source myself and it appears that most of those are checks against `None`, which seems like a bad warning rather than an actual syntax/logical issue. What did you use to get those warnings out? I can't replicate them with `-Wdefault` in a simple test file.

```
$ python -Wd <<EOF
foo = None
if foo is None:
  pass
if foo is not None:
  pass
if foo:
[...]

Issue #2693, opened 2020-02-21
Elastalert hangs during rule processing in blist module


PhracturedBlue:
I get a hang within 2-12 hours processing my rules.  Using pdb to debug is impossible because the hang appears to be in an infinite loop inside the blist module which is called with the GIL held

gdb (py-bt) show this:
```
Thread 8 (Thread 0x7fb85c229700 (LWP 3141)):
Traceback (most recent call first):         
  File "/.venv/lib/python3.7/site-packages/blist/_sortedlist.py", line 183, in __getitem__
    return self._i2u(self._blist[index])                                                                                              
  File "/.venv/lib/python3.7/site-packages/elastalert/ruletypes.py", line 336, in duration
    return self.get_ts(self.data[-1]) - self.get_ts(self.data[0])                                                                     
  File "/.venv/lib/python3.7/site-packages/elastalert/ruletypes.py", line 324, in append  
    while self.duration() >= self.timeframe:                                                                                          
  File "/.venv/lib/python3.7/site-packages/elastalert/ruletypes.py", line 620, in garbage_collect
    ({self.ts_field: ts}, 0)                                                                                                                 
  File "/.venv/lib/python3.7/site-packages/elastalert/elastalert.py", line 908, in run_rule      
    rule['type'].garbage_collect(endtime)                                                                                                    
  File "/.venv/lib/python3.7/site-packages/elastalert/elastalert.py", line 1269, in handle_rule_execution
    num_matches = self.run_rule(rule, endtime, rule.get('initial_starttime'))                                                                        
  File "/.venv/lib/python3.7/site-packages/apscheduler/executors/base.py", line 125, in run_job          
    retval = job.func(*job.args, **job.kwargs)                                                                                                       
  File "/opt/python-3.7.0/lib/python3.7/concurrent/futures/thread.py", line 57, in run                                                               
    result = self.fn(*self.args, **self.kwargs)                                                                                                      
  File "/opt/python-3.7.0/lib/python3.7/concurrent/futures/thread.py", line 80, in _worker                                                           
    work_item.run()                                                                                                                                  
  File "/opt/python-3.7.0/lib/python3.7/threading.py", line 865, in run                                                                              
    self._target(*self._args, **self._kwargs)                                                                                                        
  File "/opt/python-3.7.0/lib/python3.7/threading.py", line 917, in _bootstrap_inner                                                                 
    self.run()                                                                                                                                       
[...]

PR #3019, opened 2020-10-31
Change Library blist to sortedcontainers

nsano-rururu:
I've found that blist hasn't been updated for quite some time and that it doesn't work properly in Python 3.9, 
so I suggest changing it to another library.

Related issue
https://github.com/Yelp/elastalert/issues/2983

### DIFF
--- a/elastalert/ruletypes.py
+++ b/elastalert/ruletypes.py
@@ -3,7 +3,7 @@ import copy
 import datetime
 import sys
 
-from sortedcontainers import SortedKeyList
+from sortedcontainers import SortedKeyList as sortedlist
 
 from .util import add_raw_postfix
 from .util import dt_to_ts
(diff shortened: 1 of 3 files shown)


nsano-rururu:
There seems to be a problem with blist even in python 3.7
https://github.com/Yelp/elastalert/issues/2693

marc-:
Didn't have any issues on 3.7 or 3.8 . Popped up only in 3.9 on Debian and Alpine linux . On Alpine downgrading to 3.8 solves it.

AnthraX1:
Can someone please merge this in and push to pypi? It's currently broken on Fedora 33 with python 3.9

nsano-rururu:
@AnthraX1 

Elast Alert is no longer properly maintained. I don't think there is a pypi release.
I think I have to clone the git repository, modify the source, and do it with python setup.py install.

PR #2783, opened 2020-05-05
Fix LineNotify & Pagertree

nsano-rururu:
「'linenotify': alerts.LineNotifyAlerter,」 was not listed in elastalert/loader.py

Fixed the following issues

[#2516](https://github.com/Yelp/elastalert/issues/2516)
[#2307](https://github.com/Yelp/elastalert/issues/2307)

「'pagertree': alerts.PagerTreeAlerter,」was not listed in elastalert/loader.py

https://github.com/Yelp/elastalert/issues/2571

PR #3044, opened 2020-11-19
fix: could not import module linenotify

jweiwu:
To include LineNotifyAlerter in alerts_mapping.


nsano-rururu:
I also made a pull request for the same fix a while back
https://github.com/Yelp/elastalert/pull/2783

Issue #3041, opened 2020-11-16
Spike alert not using specified timeframe for querying event counts


piehld:
When running a spike rule with the following configuration (and verbose stdout), the time window over which elastalert is counting events appears to be that of the `run_every` time interval as opposed to the `timeframe` interval.

**Global configuration: config.yaml**
```
rules_folder: /home/ubuntu/elastalert/rules
run_every:
  minutes: 5
buffer_time:
  hours: 1
es_host: localhost
es_port: 9200
use_ssl: False
verify_certs: True
writeback_index: elastalert
writeback_alias: elastalert_alerts
alert_time_limit:
  days: 2
```

**Spike rule configuration: web-traffic-spike.yaml**
```
name: "Top-level LB stability - Dramatic change in web traffic"
es_host: my-elasticsearch-host.domain.org
es_port: 9200
index: my-index-*
type: spike
buffer_time:
  hours: 3
timeframe:
  hours: 3
spike_height: 2.5
spike_type: both
threshold_ref: 75000
# threshold_cur:
query_key: service.name
alert_on_new_data: False
use_count_query: True
doc_type: _doc
realert:
  hours: 3
alert:
- "email"
email:
- "my_email@domain.org"
from_addr: "it-services@domain.org"
```

After allowing this to run for 3 days (thus clearly surpassing the initial timeframe duration), my stdout is still reporting as follows:
```
INFO:elastalert:Background configuration change check run at 2020-11-16 07:21 PST
INFO:elastalert:Background alerts thread 0 pending alerts sent at 2020-11-16 07:21 PST
INFO:elastalert:Queried rule Top-level LB stability - Dramatic change in web traffic from 2020-11-16 07:16 PST to 2020-11-16 07:21 PST: 5661 hits
INFO:elastalert:Ran Top-level LB stability - Dramatic change in web traffic from 2020-11-16 07:16 PST to 2020-11-16 07:21 PST: 5661 query hits (0 already seen), 0 matches, 0 alerts sent
INFO:elastalert:Disabled rules are: []
INFO:elastalert:Sleeping for 299.999776 seconds
```

As you can see, the queries are only being performed over 5 min intervals (as defined for `run_every`), not 3 hours as specified for `timeframe`. Any idea on why this is happening?

nsano-rururu:
We also recommend asking questions on the Gitter channel
https://gitter.im/Yelp/elastalert

piehld:
[...]

PR #3141, opened 2021-03-02
use rule timeframe when scan_entire_timeframe is set

findmyname666:
#3041 There is rule option [scan_entire_timeframe](https://elastalert.readthedocs.io/en/latest/ruletypes.html#scan-entire-timeframe) which indicate that elastalert should scan entire timeframe set in rule. Unfortunately it scans the timeframe only in the first run. Currently it is almost impossible to ensure that rule timeframe is scan. This brings issues where errors could be overlooked. For example the elastalert query data only for last minute when *run_every* is set to 1 minute but the logs are not ingested in such fast manner. This change only affects rules with option *use_count_query* or *use_term_query*.


nsano-rururu:
@findmyname666

PR #3252, opened 2022-04-28
Fix documentation, add build scripts, etc

ghost:


nsano-rururu:
I think that it can be supported by simply upgrading the version of python. You should understand that this is not the case. I don't understand why I don't understand.

nsano-rururu:
Why is the "s" in Elasticsearch changed to uppercase even though it is lowercase? Impossible.

nsano-rururu:
I've deleted the blist, but it's different. If you don't change the library, the process won't work. Impossible response

nsano-rururu:
This change should never be merged.

nsano-rururu:
Please close.

nsano-rururu:
Please close.

nsano-rururu:
Please close.

nsano-rururu:
Please close.

nsano-rururu:
Please close.

